PR #11, opened 2021-02-16
feat: add Postgres data source

nick-zrymiak:


connector-agent/src/data_sources/postgres.rs
```diff
@@ -1,15 +1,133 @@
...
+use r2d2::{Pool, PooledConnection};
+use r2d2_postgres::{postgres::NoTls, PostgresConnectionManager};
+use std::io::Read;
+
+type PgManager = PostgresConnectionManager<NoTls>;
+type PgConn = PooledConnection<PgManager>;
+
+pub struct PostgresDataSourceBuilder {
+    pool: Pool<PgManager>,
+    data_order: Option<DataOrder>,
```
wangxiaoying, line 15: Is this data_order used anywhere? If not maybe we can remove it?

dovahcrow:
I added postgres to the CI. @nick-zrymiak can you add a script to seed the database before any postgres test run?

connector-agent/src/data_sources/postgres.rs
nick-zrymiak: Thanks Xiaoying! Will look into this.

dovahcrow (approved):

PR #14, opened 2021-02-24
Implement Pandas Writer

dovahcrow:


dovahcrow:
Fixes #16

connector-agent-python/benches/bench_string.rs
```diff
@@ -0,0 +1,136 @@
+// #![feature(custom_test_frameworks)]
```
wangxiaoying, line 1: should we delete this file or uncomment it?

dovahcrow: Let me add one more commit to remove it. I intend to keep it in the git history in case in the future we want to redo the experiment and implement the optimization.

Issue #18, opened 2021-02-26
Allow user to input date and datetime format for the schema.


dovahcrow:
The proposed syntax would be date(%Y-%m-%d)

Issue #15, opened 2021-02-24
Output None for null string


dovahcrow:
Currently, csv parser gives you an empty string.

wangxiaoying:
```
tpch=# select id, content, case when content is NULL then 1 else 0 end as isnull,
case when content = '' then 1 else 0 end as isempty,
case when content = ' ' then 1 else 0 end as isblank from test;
 id | content | isnull | isempty | isblank
----+---------+--------+---------+---------
  1 |         |      1 |       0 |       0
  2 |         |      0 |       1 |       0
  3 |         |      0 |       0 |       1
  4 | test    |      0 |       0 |       0
(4 rows)

tpch=# COPY (select * from test) TO STDOUT WITH CSV;
1,
2,""
3,
4,test
```

PR #20, opened 2021-02-27
Binary postgres

### LINKED ISSUE #19: header information is missing when dispatching
maybe we should add a method to the source called headers.

dovahcrow:
closes #19, #15, #18


connector-agent/src/data_sources/csv.rs
```diff
@@ -26,21 +35,56 @@ impl SourceBuilder for CSVSourceBuilder {
...
+    fn set_queries<Q: AsRef<str>>(&mut self, queries: &[Q]) {
+        self.files = queries
+            .into_iter()
+            .map(|fname| fname.as_ref().to_string())
+            .collect();
+    }
+
+    fn fetch_metadata(&mut self) -> Result<()> {
+        let mut reader = csv::ReaderBuilder::new()
+            .has_headers(true)
```
wangxiaoying, line 47: We assume csv file has header here, but assume it does not when do prepare. Should we let the user input whether the csv file contain the header and let the source have this information?

connector-agent/src/data_sources/csv.rs
```diff
@@ -74,136 +119,129 @@ impl DataSource for CSVSource {
...
 
-impl Produce<f64> for CSVSource {
+impl<'a> Produce<f64> for CSVSourceParser<'a> {
     fn produce(&mut self) -> Result<f64> {
-        let v: &str = self.records[self.counter / self.ncols][self.counter % self.ncols].as_ref();
-        self.counter += 1;
-        Ok(v.parse().unwrap_or_default())
+        let dt = self.schema[*self.counter % self.ncols];
+
+        if matches!(dt, DataType::F64(false)) {
```
wangxiaoying, line 185: Some types are checking type before parse and some types don't.

connector-agent/src/data_sources/csv.rs
```diff
@@ -74,136 +119,129 @@ impl DataSource for CSVSource {
...
-impl Produce<Option<i64>> for CSVSource {
+impl<'a> Produce<Option<i64>> for CSVSourceParser<'a> {
     fn produce(&mut self) -> Result<Option<i64>> {
-        let v: &str = self.records[self.counter / self.ncols][self.counter % self.ncols].as_ref();
-        self.counter += 1;
-        if v.is_empty() {
-            return Ok(None);
+        let dt = self.schema[*self.counter % self.ncols];
+
+        if matches!(dt, DataType::I64(_)) {
```
wangxiaoying, line 169: Should it be I64(true)?

connector-agent/src/data_sources/csv.rs
dovahcrow: let's first assume they all have headers. Once we want to make csv source production-ready we can add more parameters.

PR #34, opened 2021-03-03
feat: add read sql function

Wukkkinz-0725:


connector-agent-python/src/lib.rs
```diff
@@ -72,3 +72,40 @@ fn read_pg(py: Python, conn: &str, sqls: Vec<String>, schema: &str) -> PyResult<
...
+    let mut queries: Vec<String> = vec![];
+    let partition_size = match (max - min + 1) % num == 0 {
+        true => (max - min + 1) / num,
+        false => (max - min + 1) / num + 1,
+    };
+
+    for i in 0..num {
+        let lower = min + i * partition_size;
+        let upper = min + (i + 1) * partition_size;
+        let partition_query = format!(
```
wangxiaoying, line 96: Input queries may have predicates. For example, if we have a user input query: `SELECT * FROM r WHERE a > 10`, then the partition queries will become: `SELECT * FROM r WHERE a > 10 WHERE col >= lower AND  col < upper`. Maybe you can use `sqlparse` to parse the query first and then modify the query (like in `connector-agent/src/data_sources/postgres/sql.rs`) by adding additional selection

nick-zrymiak: We'll look into this.

Wukkkinz-0725: Yup, it makes sense~

nick-zrymiak: Actually just tried to test this query with some values: SELECT * FROM r WHERE a > 10

This is the output: ["SELECT * FROM r WHERE a > 10 where id >= 5 and id < 9", "SELECT * FROM r WHERE a > 10 where id >= 9 and id < 13", "SELECT * FROM r WHERE a > 10 where id >= 13 and id < 17"]

So I think it might work? 🤔

connector-agent-python/src/lib.rs
```diff
@@ -72,3 +72,40 @@ fn read_pg(py: Python, conn: &str, sqls: Vec<String>, schema: &str) -> PyResult<
         .collect();
     PyResult::Ok(ret.into_py_dict(py).to_object(py))
 }
+
+#[pyfunction]
+fn read_sql<'a>(
+    py: Python<'a>,
+    conn: &str,
+    query: &str,
+    col: &str,
```
dovahcrow, line 81: Can we make the partition optional? i.e. pass in a tuple or just first create a struct called `PartitionConfig` and derive IntoPyDict for it. for `read_sql` it takes py, conn, query and Option<PartitionConfig> and return_type. If the partition is not specified, we do not do partition.

dovahcrow: Also I feel like we need to have a wrapper on the python side to bundle the params and make some of them keyword only to make the intention of the function clear.

connector-agent-python/src/lib.rs
```diff
@@ -72,3 +72,40 @@ fn read_pg(py: Python, conn: &str, sqls: Vec<String>, schema: &str) -> PyResult<
...
+        let partition_query = format!(
+            "{} where {} >= {} and {} < {}",
+            query, col, lower, col, upper
+        );
+        queries.push(partition_query);
+    }
+    let queries: Vec<_> = queries.iter().map(|s| s.as_str()).collect();
+    match return_type {
+        "pandas" => Ok(crate::pandas::write_pandas(py, conn, &queries, false)?),
+        "arrow" => todo!(),
```
dovahcrow, line 105: I think for this we can now throw a python exception indicating this not implemented.

dovahcrow (changes requested):

connector-agent-python/src/lib.rs
```diff
@@ -72,3 +76,78 @@ fn read_pg(py: Python, conn: &str, sqls: Vec<String>, schema: &str) -> PyResult<
...
+    let mut ast = Parser::parse_sql(&dialect, query).unwrap();
+
+    match &mut ast[0] {
+        Statement::Query(q) => match &mut q.body {
+            SetExpr::Select(select) => {
+                let cur_selection = select.selection.as_ref();
+                let mut _partition_query = format!("{} >= {} and {} < {}", col, lower, col, upper);
+                if !cur_selection.is_none() {
+                    _partition_query = format!(
+                        "{} and {} >= {} and {} < {}",
```
wangxiaoying, line 94: Maybe we need to add bracket here. If the query is: `SELECT * FROM table WHERE a > 0 OR b < 10`, then this will make the partition query be: `SELECT * FROM table WHERE a > 0 OR b < 10 AND col >= upper AND col < lower`, which will be the same with `(a > 0) OR (b < 10 AND col >= upper AND col < lower)` since `And` has precedence over `Or` in sql. However what we actually want here is `(a > 0 OR b < 10) AND (col >= upper AND col < lower)`.

PR #56, opened 2021-03-28
add example to docstring

Wukkkinz-0725:


wangxiaoying:
Great! Here are a few things that would be better to add:
1. A simple description about how we do the partition in the documentation
2. A sentence about what kind of query our partition method supports (select, project, join) in the documentation
3. One test case for each supported type of query (we already have select in our test, so only queries with project and join operations are needed)

wangxiaoying (changes requested):
Looks great! Here are a few things that would be better to add:

1. A simple description about how we do the partition in the documentation
2. A sentence about what kind of query our partition method supports (select, project, join) in the documentation
3. One test case for each supported type of query (we already have select in our test, so only queries with project and join operations are needed)

dovahcrow:
great job! I polished the examples and I think you can continue resolving @wangxiaoying 's comments.

dovahcrow (approved):

Issue #26, opened 2021-03-02
Add duration type when rust support it


dovahcrow:
i16, f32/i32, interval, numeric, enum, bytes, char, time, timetz, uuid,

and add test

Yizhou150:
Hi Xiaoying,I’ve just arrived in China and currently I have no VPN and can’t access discord. I haven’t finished the postgres types since last week I was busy with my course deadlines. I’m really sorry about that. I’ll try to submit a PR today for what I have now, by the end of today, which should be Wednesday 9 am in Vancouver. Is that Okay?Sorry about that.Best,Yizhou         
            
                
                    
                            
                                
                            
                            
                                yca375
                            
                    
                        
                            
                                    ***@***.***
                            
                        
                
            
        On 03/18/2021 04:19, Xiaoying Wang wrote: 
Assigned #26 to @Yizhou150.

—You are receiving this because you were assigned.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/sfu-db/connector-agent/issues/26#event-4467425702",
"url": "https://github.com/sfu-db/connector-agent/issues/26#event-4467425702",
"name": "View Issue"
},
"description": "View this Issue on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

baggiponte:
Hi! I noticed last day that the `interval` type was not (yet) supported. It might be converted to arrow's `Duration`, am I right?

```python
from __future__ import annotations

import os

import connectorx as cx
from dotenv import load_dotenv

load_dotenv()

user: str | None = os.getenv("POSTGRESQL_USER")
password: str | None = os.getenv("POSTGRESQL_PASSWORD")
host: str | None = os.getenv("POSTGRESQL_HOST")
port: str | None = os.getenv("POSTGRESQL_PORT")
database: str | None = os.getenv("POSTGRESQL_DATABASE")

connection_uri = f"postgres://{user}:{password}@{host}:{port}/{database}"

# get top 10 values from `durata_noleggio` col which is an `interval` type
cx.read_sql(connection_uri, "SELECT durata_noleggio FROM bikemi_data LIMIT 10")
```

[...]

PR #71, opened 2021-04-18
Postgres add support for more types

wangxiaoying:
#26 

add support: i16, enum, bytea, json, jsonb, time, char, uuid
not support: timetz, interval (rust-postgres does not support)


connectorx/src/sources/postgres/mod.rs
```diff
@@ -17,13 +17,14 @@ use postgres::{
...
 use rust_decimal::Decimal;
+use serde_json::{from_str, Value};
 use sql::{count_query, get_limit, limit1_query};
 use std::marker::PhantomData;
 pub use typesystem::PostgresTypeSystem;
 use uuid::Uuid;
-
 type PgManager = PostgresConnectionManager<NoTls>;
 type PgConn = PooledConnection<PgManager>;
+use hex::decode;
```
dovahcrow, line 27: This line can be put above.

dovahcrow (changes requested):
This one seems to be accidentally submitted. `connectorx-python/connectorx/tests/tmp`

PR #82, opened 2021-06-16
Mysql

Wukkkinz-0725:


connectorx-python/connectorx/tests/test_read_sql.py
```diff
@@ -13,6 +13,25 @@ def postgres_url() -> str:
     return conn
 
 
+@pytest.fixture(scope="module")  # type: ignore
```
wangxiaoying, line 16: maybe we can make mysql related test into a separated file? test_read_sql is more like using postgres to test more common functionalities like partition, supported query types.

connectorx/src/sources/mysql/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+pub struct MysqlSourcePartition {
+    conn: MysqlConn,
+    query: String,
+    schema: Vec<MysqlTypeSystem>,
+    nrows: usize,
+    ncols: usize,
+    buf_size: usize,
+    // _protocol: PhantomData<P>,
```
wangxiaoying, line 144: can delete this comment

connectorx/src/sources/mysql/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+impl SourcePartition for MysqlSourcePartition {
+    type TypeSystem = MysqlTypeSystem;
+    type Parser<'a> = MysqlSourcePartitionParser<'a>;
+
+    fn prepare(&mut self) -> Result<()> {
+        self.nrows = match get_limit(&self.query)? {
+            // now get_limit using PostgreDialect
+            None => {
+                let row: Option<usize> = self.conn.query_first(&count_query(&self.query)?).unwrap();
+                row.unwrap()
```
wangxiaoying, line 170: Would be better if we can try to throw an error instead of using unwrap. Same for rest of the code.

connectorx/src/sources/mysql/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            for _ in 0..self.buf_size {
+                if let Some(item) = self.iter.next() {
+                    self.rowbuf.push(item.unwrap());
+                } else {
+                    break;
+                }
+            }
+
+            if self.rowbuf.is_empty() {
+                throw!(anyhow!("Postgres EOF"));
```
wangxiaoying, line 236: Here should be MySQL EOF, otherwise might be misleading

connectorx-python/src/pandas/mysql_pandas.rs
```diff
@@ -0,0 +1,55 @@
...
+    mappings = {
+        { Double[f64]                => F64[f64]                | conversion all }
+        { Long[i64]                  => I64[i64]                | conversion all }
+        { Date[NaiveDate]            => DateTime[DateTime<Utc>] | conversion half }
+        { Time[NaiveTime]            => String[String]          | conversion half }
+        { Datetime[NaiveDateTime]    => DateTime[DateTime<Utc>] | conversion half }
+        { Decimal[Decimal]           => F64[f64]                | conversion half }
+        // { Char[&'r str]              => Char[&'r str]           | conversion none }
+        // { VarChar[&'r str]           => Str[&'r str]            | conversion none }
+        { Char[String]              => Char[String]           | conversion none }
```
wangxiaoying, line 27: seems like the destination type does not match?
Char -> char
Str -> &str (maybe we should use String[String] here)

wangxiaoying (changes requested):
some .DS_Store files are accidentally submitted

connectorx/src/sources/mysql/mod.rs
Wukkkinz-0725: `    fn prepare(&mut self) -> Result<()> {
        self.nrows = match get_limit(&self.query)? {
            // now get_limit using PostgreDialect
            None => {
                let row: Option<usize> = self.conn.query_first(&count_query(&self.query)?).unwrap();
                row.unwrap_or_else(|| panic!("failed to get the count of query: {:?}", self.query))
            }
            Some(n) => n,
        };
        Ok(())
    }`
I changed the code like this, I am not sure if it is suitable or not.

Wukkkinz-0725: fn prepare(&mut self) -> Result<()> {
        self.nrows = match get_limit(&self.query)? {
            // now get_limit using PostgreDialect
            None => {
                let row: Option<usize> = self.conn.query_first(&count_query(&self.query)?).unwrap();
                row.unwrap_or_else(|| panic!("failed to get the count of query: {:?}", self.query))
            }
            Some(n) => n,
        };
        Ok(())
    }

I changed the code like this, using `unwrap_or_else`. Does that make sense or not?

PR #84, opened 2021-06-21
Add sqlite data source

wangxiaoying:
1. add sqlite source
2. reorganize code related to sql generation and source preparation
3. change string buffer size from 16MB to 4MB, and try to acquire lock when exceed 2MB


connectorx-python/src/pandas/mod.rs
```diff
@@ -1,52 +1,74 @@
...
 };
 use fehler::throws;
 use log::debug;
 use pyo3::{PyAny, Python};
 
 #[throws(ConnectorAgentPythonError)]
-pub fn write_pandas<'a>(py: Python<'a>, conn: &str, queries: &[&str], protocol: &str) -> &'a PyAny {
+pub fn write_pandas<'a>(
+    py: Python<'a>,
+    source_type: &SourceType,
```
dovahcrow, line 27: Can we merge source_type with conn? e.g. use a url parser to parse the conn string.
Also Postgres comes with two schemas: `postgres://` and `postgresql://`.

Justfile
```diff
@@ -17,15 +17,16 @@ build-python-extention:
 setup-python: build-python-extention
     cd connectorx-python && poetry run python ../scripts/python-helper.py copy-extension
     
-test-python: setup-python
-    cd connectorx-python && poetry run pytest connectorx/tests -v -s
+test-python opts="": setup-python
```
dovahcrow, line 20: `+opts=""` so you can input an arbitrary number of parameters after this command.

PR #89, opened 2021-07-05
Redirect Mysql PR

wangxiaoying:
#88 
since I don't have the permission directly modify on the original forked repo


wangxiaoying:
Also added Postgres non-copy protocol, in order to support sources like redshift

PR #90, opened 2021-07-09
arrow use largestring and date32/date64

ritchie46:
This PR proposes using `LargeString` in favor of `String` type. The `String` datatype offsets the bytes in the `&str` buffer with an `i32` meaning that the columns can only hold 2^32 == ~2GB of string data. This limit is very easily met.

The `LargeString` uses an `i64` as offset type. A limit I don't see reached any time soon. :)

This also adds `Date32` and `Date64` for the naive dates/datetimes.

What would be needed next to make this work? I assume some python integration. And from the Rust side?

### DIFF
--- a/connectorx/src/constants.rs
+++ b/connectorx/src/constants.rs
@@ -1,0 +1,1 @@
+pub(crate) const SECONDS_IN_DAY: i64 = 86_400;
--- a/connectorx/src/destinations/arrow/arrow_assoc.rs
+++ b/connectorx/src/destinations/arrow/arrow_assoc.rs
@@ -1,10 +1,12 @@
+use crate::constants::SECONDS_IN_DAY;
 use crate::errors::{ConnectorAgentError, Result};
 use arrow::array::{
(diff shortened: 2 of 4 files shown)


dovahcrow:
The code looks good to me! I think a natural next step is letting Arrow use a specialized type system as defined here https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/pandas/types.rs#L20 as well as providing conversion rules for Postgres / SQLite like here https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/pandas/transports/postgres.rs for pandas. Arrow uses the dummy type system because I was mainly using that for PoC at the beginning of this project.  

The type system and type conversion part are quite macro-heavy. Just let me know if you need some help with this.

BTW are you going to add more code into this PR or create a new one?

dovahcrow (approved):

ritchie46:
> The type system and type conversion part are quite macro-heavy. Just let me know if you need some help with this.

Ok, I will take a look. Probably somewhere in next week.

> BTW are you going to add more code into this PR or create a new one?

Nope, this one is free to merge. :)

PR #91, opened 2021-07-09
Clippy

ritchie46:
This PR fixes the clippy warnings on a run with the default features. 

I also added a clippy check in CI.


dovahcrow:
Thanks @ritchie46! I was thinking "hey we can add the Clippy check into the CI" when reading your first commit then I found your second commit :)

Adding Clippy is something I want to do but haven't got time to actually do it.

PR #92, opened 2021-07-10
Init arrow type system and use that in postgres transport

ritchie46:
This PR adds an arrow types system and uses that type system in `PostgresArrowTransport`.

I just pleased the type system here, so please take a good look. :)

Some questions:

* what does the this [bool]( https://github.com/sfu-db/connector-x/blob/3c56c539d29bc252d205a20700edd63efec301ab/connectorx-python/src/pandas/types.rs#L8) do?

A premature optimization, but if I understand correctly we can also implement `ArrowAssoc` for `&str` and `Option<&str>`, add that to the type system and thereby reduce an allocation of `String`?

### DIFF
--- a/connectorx/src/destinations/arrow/arrow_assoc.rs
+++ b/connectorx/src/destinations/arrow/arrow_assoc.rs
@@ -1,8 +1,8 @@
 use crate::constants::SECONDS_IN_DAY;
 use crate::errors::{ConnectorAgentError, Result};
 use arrow::array::{
-    ArrayBuilder, BooleanBuilder, Date32Builder, Date64Builder, Float64Builder, Int32Builder,
-    Int64Builder, LargeStringBuilder,
+    ArrayBuilder, BooleanBuilder, Date32Builder, Date64Builder, Float32Builder, Float64Builder,
+    Int32Builder, Int64Builder, LargeStringBuilder, UInt32Builder, UInt64Builder,
 };
 use arrow::datatypes::Field;
 use arrow::datatypes::{DataType as ArrowDataType, DateUnit};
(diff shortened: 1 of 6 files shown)


connectorx/src/transports/postgres_arrow.rs
```diff
@@ -9,24 +8,23 @@ pub struct PostgresArrowTransport;
...
-        { Date[NaiveDate]            => DateTime[DateTime<Utc>] | conversion half }
-        { UUID[Uuid]                 => String[String]          | conversion half }
-        { Char[&'r str]              => String[String]          | conversion none}
-        // { Time[NaiveTime]            => String[String]          | conversion half }
+        { Float4[f32]                => Float32[f32]            | conversion all }
+        { Float8[f64]                => Float64[f64]            | conversion all }
+        { Int2[i16]                  => Int32[i32]              | conversion all }
+        { Int4[i32]                  => Int32[i32]              | conversion all }
+        { Int8[i64]                  => Int64[i64]              | conversion all }
+        { Bool[bool]                 => Boolean[bool]           | conversion all  }
```
dovahcrow, line 19: There are three conversion rule generating options: all, half and none.

* `all` means the `TypeConversion<T,U>` can be done trivially using rust's `as`, e.g. `f32` to `f32` or `f32` to `f64`.
* `half` means we'd like to manually implement the `TypeConversion<T,U>` for non-optional type, e.g. `i64` to `Date64` and auto generate the type conversion of `TypeConversion<Option<T>,Option<U>>` by  `t.map(TypeConversion::convert)`.
* `none` means we'd like to manually implement the `TypeConversion<T,U>` and `TypeConversion<Option<T>, Option<U>>`, which should be rarely used.

connectorx/src/transports/postgres_arrow.rs
```diff
@@ -48,6 +46,24 @@ impl<'r> TypeConversion<&'r str, String> for PostgresArrowTransport {
     }
 }
 
+impl<'r> TypeConversion<&'r str, &'r str> for PostgresArrowTransport {
```
dovahcrow, line 49: Following the comment above, I think this impl can be replaced by an `all` rule.

connectorx/src/transports/postgres_arrow.rs
ritchie46: Thanks for the explanation. :+1: Makes more sense now.

connectorx/src/destinations/arrow/arrow_assoc.rs
```diff
@@ -18,6 +18,74 @@ pub trait ArrowAssoc {
...
+        UInt32Builder::new(nrows)
+    }
+
+    #[throws(ConnectorAgentError)]
+    fn append(builder: &mut UInt32Builder, value: u32) {
+        builder.append_value(value)?;
+    }
+
+    fn field(header: &str) -> Field {
+        Field::new(header, ArrowDataType::UInt64, false)
```
dovahcrow, line 34: Should this be UInt32?

connectorx/src/destinations/arrow/arrow_assoc.rs
```diff
@@ -86,6 +154,40 @@ impl ArrowAssoc for Option<i64> {
...
+        Float32Builder::new(nrows)
+    }
+
+    #[throws(ConnectorAgentError)]
+    fn append(builder: &mut Self::Builder, value: f32) {
+        builder.append_value(value)?;
+    }
+
+    fn field(header: &str) -> Field {
+        Field::new(header, ArrowDataType::Float64, false)
```
dovahcrow, line 170: Should this be Float32?

dovahcrow:
Hi @ritchie46 , I added a test that copies data from postgres to arrow, as well as some minor refactors. Let's merge the PR if you are ok with the change.

ritchie46:
> Hi @ritchie46 , I added a test that copies data from postgres to arrow, as well as some minor refactors. Let's merge the PR if you are ok with the change.

Nice, That means that the rust side of thing has a working PoC? Or is more needed? 

I think this one can get merged in. :+1:

dovahcrow:
> Nice, That means that the rust side of thing has a working PoC? 

Yes, it is a working PoC! I guess what's left is:

1. adding more types
2. finish the conversions from mysql/sqlite to arrow
3. publish it to crates.io

ritchie46:
> Yes, it is a working PoC! I guess what's left is:

    adding more types
    finish the conversions from mysql/sqlite to arrow
    publish it to crates.io
    implement the pyarrow support in python

Great! :slightly_smiling_face: 

I will take a look at the python side.

Issue #103, opened 2021-07-22
Postgres SSL Support


wseaton:
Currently no TLS implementation is supported, so connecting like this fails if the DB mandates an SSL/TLS connection:

`df = cx.read_sql(f"postgresql://{username}:{password}@host:35432/schema?sslmode=require", 'select 1;', return_type='pandas')`

Error:
```
RuntimeError: timed out waiting for connection: error performing TLS handshake: no TLS implementation configured
```

Is this a planned feature? It seems to be supported upstream in `rust-postgres`: https://docs.rs/postgres-native-tls/0.5.0/postgres_native_tls/

The lack of SSL support (even with `sslmode=require`) makes it difficult to use connector-x in enterprise-y environments.

dovahcrow:
Thanks for the reminder on SSL! I almost forgot this important feature. Will add this to the todo list for v0.2.0

PR #104, opened 2021-07-24
Add SSL/TLS support for Postgres connections

wseaton:
As discussed in #103, currently there's no way to connect to a postgres database that mandates a TLS/SSL connection.

What this PR does is make it so `sslmode=required` is functional, and if a root cert and no client certs are passed it falls back to the behavior of `sslmode=verify-ca`. 

Example conn string:
```python
conn = f"postgresql://{user}:{pw}@my.database.host:5439/db?sslmode=require&sslrootcert=/tmp/myroot.crt"
df = cx.read_sql(conn, "select 1;", return_type='pandas', protocol='cursor')
```

There is a lot of upstream discussion on the topic and I've tried to link to it in doc strings where appropriate. Tested it on Amazon Redshift, but looking for pointers on how to add functional tests for this. All of the code probably isn't idiomatic rust either, so looking for some feedback there too. Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -362,13 +362,13 @@ dependencies = [
  "iai",
  "itertools 0.10.1",
  "log",
+ "native-tls",
  "ndarray",
  "num-traits",
- "openssl",
  "owning_ref",
  "polars",
  "postgres",
- "postgres-openssl",
+ "postgres-native-tls",
  "pprof",
  "r2d2",
  "r2d2_mysql",
@@ -407,6 +407,7 @@ dependencies = [
  "ndarray",
  "numpy",
  "postgres",
+ "postgres-native-tls",
  "pprof",
  "pyo3",
  "pyo3-built",
@@ -2002,15 +2003,15 @@ dependencies = [
 ]
 
 [[package]]
-name = "postgres-openssl"
+name = "postgres-native-tls"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1de0ea6504e07ca78355a6fb88ad0f36cafe9e696cbc6717f16a207f3a60be72"
+checksum = "2d442770e2b1e244bb5eb03b31c79b65bb2568f413b899eaba850fa945a65954"
 dependencies = [
  "futures",
- "openssl",
+ "native-tls",
  "tokio",
- "tokio-openssl",
+ "tokio-native-tls",
  "tokio-postgres",
 ]
 
@@ -2962,14 +2963,12 @@ dependencies = [
 ]
 
 [[package]]
-name = "tokio-openssl"
-version = "0.6.2"
+name = "tokio-native-tls"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f24cddc8445a4dc8359cdd9e91c19d544fc95f672e32afe8945852b9381a09fe"
+checksum = "f7d995660bd2b7f8c1568414c1126076c13fbb725c40112dc0120b78eb9b717b"
 dependencies = [
- "futures",
- "openssl",
- "openssl-sys",
+ "native-tls",
  "tokio",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,16 +252,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40e38929add23cdf8a366df9b0e088953150724bcbe5fc330b0d8eb3b328eec8"
 
 [[package]]
-name = "built"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


wangxiaoying:
Hi @wseaton, thanks for the PR! The logic of the connection construction looks pretty good to me and the docstrings make it very clear. There are only two things that I think might need some changes:

1. We already have the dependency on `native-tls` (for mysql), so it would be better if we use the same for postgres
2. Before the source construction (`PostgresSource::new`), we will also query the range of the partition key in the case that partition number is given and range is not (`source_router::pg_get_partition_range`). The connection construction here also need to be modified in order to support TLS.

Maybe I can make some modifications for the 2 issues above as well as add some tests. Let me know if you have any concerns : )

wseaton:
Hey @wangxiaoying, thanks for the quick review!

>     1. We already have the dependency on `native-tls` (for mysql), so it would be better if we use the same for postgres
> 

Hmm, I agree, although I could not find a method to add custom client certificates when using `postgres-native-tls`, only the addition of a trusted root ca. In addition I think the postgres docs mention [requiring openssl for clients](https://www.postgresql.org/docs/13/ssl-tcp.html#SSL-SETUP) so it may be unavoidable, perhaps it could be gated behind a feature flag defaulting to `NoTls` if unset?

>     2. Before the source construction (`PostgresSource::new`), we will also query the range of the partition key in the case that partition number is given and range is not (`source_router::pg_get_partition_range`). The connection construction here also need to be modified in order to support TLS.
> 

I can definitely fix this one.

> 
> I can make some modifications for the 2 issues above as well as add some tests. Let me know if you have any concerns : )

Sure, feel free to make any modifications you wish :)

One other thing I noticed was that the type alias for `PgManager` is not very ergonomic due to generics and may cause issues if we want to alternate between `NoTls` and a `MakeTlsConnector` implementation, not sure what to do about that.

wangxiaoying:
@wseaton Thanks for the quick reply : )

> Hmm, I agree, although I could not find a simple method to add custom client certificates when using postgres-native-tls, only the addition of a trusted root ca. In addition I think the postgres docs mention requiring openssl for clients so it may be unavoidable, perhaps it could be gated behind a feature flag defaulting to NoTls if unset?

Yeah, I also found that `openssl` is necessary if we need to support client authentication. But it seems like it is not a common need. According to this [doc](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION), client side authentication is used to defend against impersonation, which is typically due to insecure password management. Maybe we can only support server authentication for now (`sslmode>=require` and `sslrootcert`) and add the client authentication if there are needs for this? (It seems like `sslpassword ` is also related to client certificate, we can add support for this together if needed.) 


> One other thing I noticed was that the type alias for PgManager is not very ergonomic due to generics and may cause issues if we want to alternate between NoTls and a MakeTlsConnector implementation, not sure what to do about that.

Totally agreed. I'm trying to see whether there is a good way to make it more generic. It would be better if we could alter between `NoTLS` and `MakeTlsConnector`.

wseaton:
@wangxiaoying refactored things a little bit, and moved the main logic for SSL construction into `impl From<TlsConfig> for MakeTlsConnector`, also fixed 2) (the source_router issue) from above

dovahcrow:
Hi @wseaton we made a big refactor in the last 24 hours. I rebased your code to the refactored main branch and solved several conflicts and force pushed. You can do `git fetch --all && git reset --hard origin/postgres_ssl` to get the latest code. Be sure to save all your uncommitted changes first since this command will override all the differences from your local branch with the remote one.

wseaton:
@wangxiaoying just tested your latest `postgres-native-tls` refactor and everything is looking good w/ redshift via the connection strategy I was using above. Let me know if there's anything else I can do to help!

wangxiaoying:
> @wangxiaoying just tested your latest `postgres-native-tls` refactor and everything is looking good w/ redshift via the connection strategy I was using above. Let me know if there's anything else I can do to help!

@wseaton That's great! I cleaned up the code a little bit and tested with postgres on aws rds. If you think it's ok then we can merge it after review!

frostpuppet:
Is there any possibility of adding support for the sslkey and sslcert for client auth?
Still somewhat new to rust but wondering how hard an addition like this is 
https://www.cockroachlabs.com/docs/stable/build-a-rust-app-with-cockroachdb.html

wangxiaoying:
Hi @frostpuppet , we currently only support server-side authentication since client auth is less common. Adding support for client auth may require `openssl` crate (current `native-tls` is not enough). Actually this [initial commit](https://github.com/wseaton/connector-x/blob/10cc63630c8c8c7c2dd0eea62f0dbc852dd598d3/connectorx/src/sources/postgres/connection.rs) supported it and we remove it later since we want to have less unnecessary dependencies.

May I ask what's your use case scenario for client auth?

frostpuppet:
Thanks for replying so quickly.  My use case really revolves around the way our database is secured.  Typically a client will request access and certificates are issued through a Vault.  So in my case I only have 3 certs and no password is issued.

PR #110, opened 2021-08-12
More mysql types

Yizhou150:


wangxiaoying:
The code looks good to me. Can you also update this [doc](https://github.com/sfu-db/connector-x/blob/main/Types.md) by adding the newly supported type for MySQL? I think we can merge this pr after that.

Issue #149, opened 2021-11-02
[Postgres] int4 type conversion failure on partition compare


wseaton:
Related to #107, I can get queries to run now via the postgres bridge of my database (via the `cursor` protocol) but I'm running into a subtle type issue. Was wondering if you guys had any idea the cause, happy to help contribute a fix. 

```
thread '<unnamed>' panicked at 'error retrieving column 0: error deserializing column 0: cannot convert between the Rust type `i64` and the Postgres type `int4`', /github/home/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-postgres-0.7.3/src/row.rs:151:25
```

From what I can tell in the transport mapping, this conversion shouldn't be happening (https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/postgres/typesystem.rs#L41). 

The error only happens on the `COUNT(*)` queries used to estimate the partition size. 

Example:
```
[2021-11-02T13:44:35Z DEBUG connectorx::sql] Transformed count query: SELECT count(*) FROM (SELECT cn, account, created_fiscal_yr FROM mydb.mytable WHERE createddate >= '2017-01-01') AS CXTMPTAB_PART WHERE 2019 <= CXTMPTAB_PART.created_fiscal_yr AND CXTMPTAB_PART.created_fiscal_yr < 2022
```

Maybe there is a code path there were the arrow transport is not used? Thanks!

wangxiaoying:
Hi @wseaton , thanks for the report!

There are two places that we will directly parse the result without going through type transport.

One is the count query here: https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/postgres/mod.rs#L308. It happens in both partition and no partition cases.

Another place is when we partition the query, we need to get the range (min and max) of the partition column first: https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/source_router.rs#L113. This will only happen when `partition_on` and `partition_num` is set.

Both of these two places happens before "Start writing" log.  Can you help to check whether the error is caused by the two cases above or after "Start writing" (where the later conversion will be done according to the typesystem you mentioned)?

wseaton:
@wangxiaoying thanks for the quick response! The bug definitely happens after `Prepare partitions` and before `Allocate destination memory`, and I'm hitting it while both providing and not providing a partition column to split on

wangxiaoying:
[...]

PR #153, opened 2021-11-02
Fix for #149

wseaton:
This patch solves the issue outlined in #149 on my particular database, could probably be refactored to reduce repetition but couldn't figure out how in a way I liked, any pointers? :slightly_smiling_face:

### DIFF
--- a/connectorx/src/sources/postgres/mod.rs
+++ b/connectorx/src/sources/postgres/mod.rs
@@ -202,6 +202,30 @@ where
             _protocol: PhantomData,
         }
     }
+
+    #[throws(PostgresSourceError)]
+    pub fn get_total_rows(&mut self) -> usize {
+        let dialect = PostgreSqlDialect {};
+        let row = self
+            .conn
+            .query_one(count_query(&self.query, &dialect)?.as_str(), &[])?;
+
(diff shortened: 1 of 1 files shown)


wangxiaoying:
Thanks @wseaton ! The code looks good. Can you also add this modification to the `BinaryProtocol`'s prepare stage: https://github.com/sfu-db/connector-x/pull/153/files#diff-9a26d294fe1517a58485b5b2709196a4613e51fd9f99acd82ca16028a771b53eL218 to make the count consistent? (It seems like we didn't add the check limit on `CSVProtocol` and `CursorProtocol` previously, can you also help to add this check? Thanks for helping to find this inconsistency!)

The repetition looks ok to me since it's not much. But if we want to reduce it, I think one way might be extracting a count function, which's input is reference to a connection and a query, and output is the count result in `usize`. All three protocols can invoke it in this function prepare stage and set the `self.nrows` using the returned result.

wseaton:
@wangxiaoying was above to remove the duplication as per your suggestion, still works great against my database. Thanks for the help!

wangxiaoying:
@wseaton Looks great! Thanks for the PR. I have merged this.

PR #154, opened 2021-11-03
add support for polars >= 0.8

wseaton:
Currently connectorx pins the version of polars to `0.8.*`, but it's possible to add support for later versions of polars without breaking backwards compatibility by just attempting the import and catching the exception. What do y'all think?

(cc @ritchie46)

### DIFF
--- a/connectorx-python/connectorx/__init__.py
+++ b/connectorx-python/connectorx/__init__.py
@@ -146,7 +146,11 @@ def read_sql(
             except ModuleNotFoundError:
                 raise ValueError("You need to install polars first")
 
-            df = pl.DataFrame.from_arrow(df)
+            try: 
+                df = pl.DataFrame.from_arrow(df)
+            except AttributeError:
+                # api change for polars >= 0.8.*
+                df = pl.from_arrow(df)
     else:
         raise ValueError(return_type)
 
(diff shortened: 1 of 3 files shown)


dovahcrow:
@wseaton What is the guarantee for the version number in Polars? In Semver it only guarantees no breaking changes between `0.x.y` and `0.x.z`, but not between `0.x` and `0.y`. If Polars has a different guarantee, e.g. no breaking changes until 1.0, we could definitely do `>=0.8`.

wseaton:
@dovahcrow good point, I think I'd defer to @ritchie46 on any future plans for change. The `from_arrow` API is one of the most foundational APIs of the library, so I'd hope it wouldn't be changing any more between versions and can be considered 'settled code'.

As for the brittleness of the import shim, a polars 1.0 user could just use `return_type='arrow'` as a workaround if this in the future was to ever break so there's at least an escape hatch.

dovahcrow:
> @dovahcrow good point, I think I'd defer to @ritchie46 on any future plans for change. The `from_arrow` API is one of the most foundational APIs of the library, so I'd hope it wouldn't be changing any more between versions and can be considered 'settled code'.
> 
> As for the brittleness of the import shim, a polars 1.0 user could just use `return_type='arrow'` as a workaround if this in the future was to ever break so there's at least an escape hatch.

Sounds good to me for ">=0.8"  if `from_arrow` is considered settled.

dovahcrow:
Let's merge this PR then. I think from_arrow is the only function we require from Polars.

PR #191, opened 2021-12-09
Automatically select 'text' protocol for clickhouse-prefixed connection strings

alexander-beedie:
much the same as the previous redshift patch, but this one is for ease of use with clickhouse. sits in the same block of code quite neatly, and doesn't disturb any other defaults. hopefully of similar utility ;)

`read_sql("clickhouse://...details...", query=...)` => `read_sql("mysql://...details...", query=..., protocol='text')`


alexander-beedie:
FYI: I investigated, but the current the test failures don't appear to be connected to the patch.

wangxiaoying:
Hi @alexander-beedie , I was also thinking to make clickhouse align with redshift! Thank you for the PR!

PR #228, opened 2022-02-04
Bigquery more

Yizhou150:
tests and docs


connectorx-python/connectorx/tests/test_bigquery.py
```diff
@@ -1,4 +1,5 @@
 import os
+from pickle import TRUE
```
wangxiaoying, line 2: Is this import necassary?

PR #247, opened 2022-03-03
Add documentation

Wukkkinz-0725:
Add documentation


docs/_config.yml
```diff
@@ -0,0 +1,24 @@
...
+
+# Force re-execution of notebooks on each build.
+# See https://jupyterbook.org/content/execute.html
+execute:
+  execute_notebooks: force
+
+
+# Information about where the book exists on the web
+repository:
+  url: https://github.com/Wukkkinz-0725/connector-x.git  # Online location of your book
```
wangxiaoying, line 16: should we update the url and branch here?

Wukkkinz-0725: Yes, it is my fault. I will check other part now.

Issue #281, opened 2022-05-11
Add postgres simple protocol support


wangxiaoying:
Since redshift exports data more efficiently using simple protocol than extended protocol.

PR #322, opened 2022-08-09
Add simple query

alswang18:
Added support for simple query protocol used by postgres-rust package. 

Issues: https://github.com/sfu-db/connector-x/issues/281


connectorx/src/sources/postgres/mod.rs
```diff
@@ -961,3 +966,771 @@ impl_produce!(
...
+
+                #[throws(PostgresSourceError)]
+                fn produce(&'r mut self) -> $t {
+                    let (ridx, cidx) = self.next_loc()?;
+                    let val = match &self.rows[ridx] {
+                        SimpleQueryMessage::Row(row) => match row.try_get(cidx)? {
+                            Some(s) => s
+                                .parse()
+                                .map_err(|_| ConnectorXError::cannot_produce::<$t>(Some(s.into())))?,
+                            None => panic!("get NULL for non-NULL column"),
```
wangxiaoying, line 1076: Would be better if we can throw an error instead of panic in the code (unless it's for our own testing). Can you try to throw an `cannot_produce` or `anyhow` error for all the panics?

connectorx/src/sources/postgres/mod.rs
```diff
@@ -961,3 +966,771 @@ impl_produce!(
...
+                            panic!("what?");
+                        }
+                    };
+                    val
+                }
+            }
+        )+
+    };
+}
+// macro_rules! impl_simple_vec_produce {
```
wangxiaoying, line 1116: remove this?

connectorx/src/sources/postgres/mod.rs
```diff
@@ -961,3 +966,771 @@ impl_produce!(
...
+}
+
+fn parse_input_i16(input: &str) -> Result<Vec<i16>, std::num::ParseIntError> {
+    input
+        .split(",")
+        .map(|token| token.parse())
+        .collect()
+}
+
+impl<'r> Produce<'r, Vec<i16>> for PostgresSimpleSourceParser {
```
wangxiaoying, line 1230: This looks great! Is it possible to use a macro to handle all the vectors?

wangxiaoying (changes requested):
Thanks Alec, the code looks great! I left a few comments.

A reminder is that you can use `cargo fmt` to format the code before submission. Otherwise our CI will not pass

alswang18:
> Thanks Alec, the code looks great! I left a few comments.
> 
> A reminder is that you can use `cargo fmt` to format the code before submission. Otherwise our CI will not pass

got it thanks for the reminder xiaoying :) just ran it

connectorx/src/sources/postgres/mod.rs
alswang18: Im having some issue imaging running a macro for these parsing functions

```
fn parse_input_i32(input: &str) -> Result<Vec<i32>, std::num::ParseIntError> {
    input.split(",").map(|token| token.parse()).collect()
}
```

connectorx/src/sources/postgres/mod.rs
alswang18: removed missed this

connectorx/src/sources/postgres/mod.rs
alswang18: it was for debugging, will do this shortly :)

connectorx-python/src/arrow2.rs
```diff
@@ -20,7 +38,209 @@ pub fn write_arrow<'a>(
     origin_query: Option<String>,
     queries: &[CXQuery<String>],
 ) -> &'a PyAny {
-    let destination = get_arrow2(source_conn, origin_query, queries)?;
+    let mut destination = ArrowDestination::new();
```
wangxiaoying, line 41: This part of the code is relocated [here](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/get_arrow2.rs). Can copy the newly added simple protocol related to the new location and keep this file not modified. The same for `arrow.rs`.

connectorx/src/sources/postgres/mod.rs
```diff
@@ -961,3 +965,745 @@ impl_produce!(
...
+    #[throws(PostgresSourceError)]
+    fn produce(&'r mut self) -> &'r str {
+        let (ridx, cidx) = self.next_loc()?;
+        let val = match &self.rows[ridx] {
+            SimpleQueryMessage::Row(row) => match row.try_get(cidx)? {
+                Some(s) => s,
+                None => throw!(anyhow!("Cannot parse NULL in non-NULL column.")),
+            },
+            SimpleQueryMessage::CommandComplete(c) => {
+                panic!("get command: {}", c);
```
wangxiaoying, line 1135: Also remember to throw error instead of panic here and below (with a better error message for example "got unknown simple query message ...").

wangxiaoying (changes requested):
Thanks for the update! Just note what we discussed and another minor issue : )

PR #346, opened 2022-09-05
Adding bigquery job location in parameters while getting job details

phanindra-ramesh:


wangxiaoying:
Hi @phanindra-ramesh , thank you so much for the PR! Other than the `bigquery/mod.rs` file, we also issue query to bigquery to get some metadata from it [here](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/source_router.rs#L430). Do you think this part should also add the location?

phanindra-ramesh:
hi @wangxiaoying , adding location in the QueryRequest is not required [`here`](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/source_router.rs#L430) the specified code snippet makes a call to bigquery API to get the query resultset where as, the code where I made changes requests for the result set for the job that was already submitted in the previous step.

wangxiaoying:
Hi @phanindra-ramesh , got it. I have merged the code and released an alpha version for this feature. Can you try `pip install connectorx==0.3.1a6` to test whether it works?

PR #364, opened 2022-09-25
bump arrow-rs to 22

houqp:


wangxiaoying:
Thanks for the PR! Seems like [this](https://github.com/sfu-db/connector-x/pull/364/files#diff-4dc77aaed9f964789b5e523dd328c8e1639c7a7619afcc366bfd8c1819b12254R54) is introduced by the new `datafusion` version where the result should be `NULL` instead of `0`. Taking a note here just in case.

Issue #381, opened 2022-10-24
Support for Postgres type ltree


auyer:
The Postgres type ltree is not supported. 

It is available in the client since this :  https://github.com/sfackler/rust-postgres/issues/389

I am interested in contributing this to the project.

PR #382, opened 2022-10-24
[WIP] Adds ltree string type

auyer:
This adds ltree type as a string, using commit as guide 9ae1ecb644742c6aaa29c2f0a0f1e3bd1219bae5. 
ltree is available in the driver since https://github.com/sfackler/rust-postgres/issues/389
Works with all protocols, but has an issue with Binary:

image

When I inspect the data in the shell, the dataframe that came from Postgres shows this hex prefix.
image

I'm looking for help delivering this, and other Postgres Types. Thanks!
Fixes: #381

### DIFF
--- a/connectorx-python/connectorx/tests/test_postgres.py
+++ b/connectorx-python/connectorx/tests/test_postgres.py
@@ -450,7 +450,7 @@ def test_postgres_with_index_col(postgres_url: str) -> None:
 
 
 def test_postgres_types_binary(postgres_url: str) -> None:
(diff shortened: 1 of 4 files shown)


wangxiaoying:
Hi @auyer , thanks for the PR and it looks great!

For the error, I tried to print output value from `rust-postgres` directly and it shows the return value using `copy_out` with `BINARY` has this single hex char prefix for `ltree` type:

image

which does not happen in `copy_our` with `CSV` or `query_raw` in [rust-postgres](https://github.com/sfackler/rust-postgres):

image

I think this is an issue from `rust-postgres`, and we can report this to them. In the meanwhile, we can remove this `ltree` type test from `test_postgres_types_binary` for now and add a comment in our document of the [postgres type mapping](https://sfu-db.github.io/connector-x/databases/postgres.html) (which corresponds to the markdown file at `/docs/databases/postgres.md#postgres-pandas-type-mapping`).

auyer:
Sounds great ! 
I will open the issue there and make the changes to this PR.
Tomorrow I plan to start work on Arrays ! Any advice is welcome :grin:

PR #383, opened 2022-10-26
trivial fix: typos

alexander-beedie:
I just happened to notice these while I was reading through the source (to check for PostgreSQL arrow2 array support - which now exists!) Thanks for the great work :)

PS: any (approximate) date in mind for a 0.3.1 release?

### DIFF
--- a/Types.md
+++ b/Types.md
@@ -1,6 +1,6 @@
 # Supported Protocols, Data Types and Mappings
 
-**Currently we assume all columns are nullabel when inferring schema.**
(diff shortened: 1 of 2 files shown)


wangxiaoying:
Hi @alexander-beedie , thanks for the PR!

> any (approximate) date in mind for a 0.3.1 release?
We are going to publish a stable release by this month. In the mean while, please feel free to use our latest alpha release `0.3.1a9` from pypi.

PR #392, opened 2022-11-01
Adds support for the rest of the Ltree family + docs on binary protocol

auyer:
Adds: lquery, ltxtquery 
Binary protocol for ltree types will fail in postgres older than 13


auyer:
About the version problem with ltree binary: https://www.postgresql.org/message-id/flat/CANmj9Vxx50jOo1L7iSRxd142NyTz6Bdcgg7u9P3Z8o0=HGkYyQ@mail.gmail.com

Issue #218, opened 2022-01-28
Support for tokio_postgres as a source


glennpierce:
I am using tokio_postgres im my project and was experimenting creating polars dataframe
from my db using connectorx.

Calling connectorx causes errors like 

panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.

Could tokio_postgres be supported as a source ?

wangxiaoying:
Hi @glennpierce , I'm not sure whether I understand your question correctly, but can you spawn a new thread to run the dispatcher? That should solve the error you mention. For example like this:
```Rust
let records = handle
        .spawn_blocking(|| -> Vec<RecordBatch> {
            let mut destination = ArrowDestination::new();
            let dispatcher =
                Dispatcher::<_, _, PostgresArrowTransport<PgBinaryProtocol, NoTls>>::new(
                    source,
                    &mut destination,
                    queries,
                    origin_query,
                );
            dispatcher.run().unwrap();
            destination.arrow().unwrap()
        })
        .await.unwrap();
```

glennpierce:
Thanks. For future reference I did it like

```
let df = task::spawn_blocking(move || {
            let mut destination = Arrow2Destination::new();
            let dispatcher = Dispatcher::<_, _, PostgresArrow2Transport<BinaryProtocol, tokio_postgres::NoTls>>::new(
                builder,
                &mut destination,
                &queries,
                Some(sql.clone()),
            );
            dispatcher.run().unwrap();
            destination.polars().unwrap()
        })
        .await.unwrap();
```

PR #397, opened 2022-11-08
Make SourceConn clonable and bump Arrow/Datafusion

gruuya:
Hi folks 👋🏻 

We're using connector-x to build a [remote table](https://github.com/splitgraph/seafowl/pull/190) querying capability in [Seafowl](https://seafowl.io/) (an analytical, cache friendly, time travel-able database we're building on top of Datafusion).

Given that on one hand we rely on tokio async runtime, while on the other hand [some of the DB drivers](https://github.com/sfu-db/connector-x/issues/218) block the thread when executing queries, we're forced to run the queries in a closure using [`spawn_blocking`](https://docs.rs/tokio/0.2.22/tokio/task/fn.spawn_blocking.html).

In turn, instead of creating a single `SourceConn` upon instantiating the remote table, we're forced to instantiate a new `SourceConn` for each query execution,  since otherwise we can't pass the ref to the `get_arrow` function inside the closure due to lifetimes incompatibilities (`borrowed data escapes outside of associated function`).

Long story short, to mitigate this we'd really only need `SourceConn` to be cloneable (unless this is for some reason a bad idea?), as then we could get away with keeping a single one in remote table struct and clone it or it's fields when needed, hence this PR.

Additionally, I bumped Arrow and Datafusion to the latest available versions (26 and 14 respectively).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,11 +118,16 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "arrow"
-version = "22.0.0"
+version = "26.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c5936b4185aa57cb9790d8742aab22859045ce5cc6a3023796240cd101c19335"
+checksum = "e24e2bcd431a4aa0ff003fdd2dc21c78cfb42f31459c89d2312c2746fe17a5ac"
 dependencies = [
  "ahash 0.8.0",
+ "arrow-array",
+ "arrow-buffer",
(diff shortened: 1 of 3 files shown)


alswang18:
@wangxiaoying I think this should be fine. We are only making a string container clonable as far as I can tell for this PR. Would love to know your thoughts.

wangxiaoying:
Hi @gruuya , thanks for the PR! To pass the [CI](https://github.com/sfu-db/connector-x/actions/runs/3423127595/jobs/5706315619), I made a minor change to align `arrow` version in rust with cpp and python crates. I'm going to close this PR and redirect to #399 .

gruuya:
Great, thank you!

PR #405, opened 2022-11-21
Add PG JSON to Arrow transport rules

gruuya:
There are a couple of (related) changes in this PR, all of which underpin the [remote table](https://seafowl.io/docs/guides/remote-tables) functionality that we've recently built on Seafowl: 

- the JSON transport mappings just align the existing implementation in Arrow 2 with the missing one in Arrow.
- likewise, the missing arrow association rules for `DateTime<Utc>` and `Option<DateTime<Utc>>` (PG Timestampz type) are added. Unlike in Arrow 2, we don't have a builder that will collect the Timestamp values with a UTC zone set, so we have to set it to `None` instead.
- finally bump sqlite related libs, and fix resulting conflicts (`columns` and `column_names` function changes).

Basically, we use these changes on our connector-x fork, but given that they are pretty general it may be useful to contribute them upstream.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2255,9 +2255,9 @@ checksum = "348108ab3fba42ec82ff6e9564fc4ca0247bdccdc68dd8af9764bbc79c3c8ffb"
 
 [[package]]
 name = "libsqlite3-sys"
-version = "0.22.2"
+version = "0.24.2"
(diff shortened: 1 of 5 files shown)


wangxiaoying:
Hi @gruuya , thanks for the PR, it looks great!

PR #441, opened 2023-01-30
fix build only with src_bigquery and dst_arrow2

marianoguerra:
before: `cargo build --features "dst_arrow2 src_bigquery"` failed, now it works


wangxiaoying:
Thank you @marianoguerra !

Issue #469, opened 2023-03-08
Oracle column with DATE datatype loses time part


romiof:
#### What language are you using?

Using with **Python**

#### What version are you using?

0.3.1

#### What database are you using?

Oracle 12c

#### What dataframe are you using?

PolaRS with PyArrow

#### Can you describe your bug?

I've a lot of tabels in my Oracle database, which have columns with DATE datatype.

This datatype in Oracle, used to have DATE or DATE+TIME data.

When selected this columns, CX convert it to DATE and I lose the time part of the column.

#### What are the steps to reproduce the behavior?
Selecting SYSDATE from Oracle is enough to reproduce it.
```python
import connectorx as cx

ENGINE_PATH = oracle+oracledb://user:MyP4ssword@oraclesrv01.mycompany.com:1526/MyServiceName
query = "SELECT SYSDATE FROM dual"

cx.read_sql(ENGINE_PATH, query, return_type="arrow")
# also happen with return_type="polars" or "pandas"

pyarrow.Table
COL1: date32[day]
----
COL1: [[2023-03-08]]
```

If possible, please include a **minimal simple** example including:

##### Database setup if the error only happens on specific data or data type

Any table with DATE datatype.

##### Other thoughts
If I use CAST(foo AS TIMESTAMP) I can preserve the time part when load in CX.

But I'm planning to do a Extract-Load of a LOT of tables, and write all columns in my query of all tables, will be to much verbose.

romiof:
**@ alexander.beedie** in PolaRS' Discord, [helped with it](https://discord.com/channels/908022250106667068/911186243465904178/1083109762306166805):

`Yup; looking at their source code I'm afraid it looks like DATE => Date (which makes sense by name, but obviously Oracle's DATE is more of a TIMESTAMP with no fractional seconds, so...)`

```rust
impl_transport!(
    name = OracleArrow2Transport,
    error = OracleArrow2TransportError,
    systems = OracleTypeSystem => Arrow2TypeSystem,
    route = OracleSource => Arrow2Destination,
    mappings = {
        { NumFloat[f64]                 => Float64[f64]                 | conversion auto }
        { Float[f64]                    => Float64[f64]                 | conversion none }
        { BinaryFloat[f64]              => Float64[f64]                 | conversion none }
        { BinaryDouble[f64]             => Float64[f64]                 | conversion none }
        { NumInt[i64]                   => Int64[i64]                   | conversion auto }
[...]

PR #470, opened 2023-03-13
Fixed oracle date type to rust naive datetime - issue 469

FerriLuli:
Oracle date type is more like a timestamp, to fix the problems listed on the issue https://github.com/sfu-db/connector-x/issues/469 I altered the Oracle `typesystem.rs` file.

### DIFF
--- a/connectorx/src/sources/oracle/typesystem.rs
+++ b/connectorx/src/sources/oracle/typesystem.rs
@@ -14,7 +14,6 @@ pub enum OracleTypeSystem {
     Char(bool),
     NVarChar(bool),
     NChar(bool),
-    Date(bool),
     Timestamp(bool),
     TimestampTz(bool),
 }
@@ -26,7 +25,7 @@ impl_typesystem! {
         { Float | NumFloat | BinaryFloat | BinaryDouble => f64 }
         { Blob => Vec<u8>}
         { Clob | VarChar | Char | NVarChar | NChar => String }
-        { Date => NaiveDate }
+        { Date => NaiveDateTime }
         { Timestamp => NaiveDateTime }
         { TimestampTz => DateTime<Utc> }
     }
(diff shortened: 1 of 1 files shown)


connectorx/src/sources/oracle/typesystem.rs
wangxiaoying, line 1: Thanks for the PR.

Instead of removing the `OracleTypeSystem::Date` type, maybe it's better to keep it but map it to `NaiveDateTime` instead of `NaiveDate` in `impl_typesystem` in order to make the types clear for an 1-1 mapping (like VarChar, Char...)?

Another thing is that to be able to compile, you may also need to modify all the `impl_transport!`. For example, for oracle to arrow data transport in [here](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/oracle_arrow.rs#L41), update into `{ Date[NaiveDateTime]   => Date64[NaiveDateTime]      | conversion auto }` (like timestamp rows). The same applies to files including: [oracle to arrow2](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/oracle_arrow2.rs#L43), [oracle to pandas](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/pandas/transports/oracle.rs#L30).

wangxiaoying (changes requested):
Thanks for the PR.

Instead of removing the `OracleTypeSystem::Date` type, maybe it's better to keep it but map it to `NaiveDateTime` instead of `NaiveDate` in `impl_typesystem` in order to make the types clear for an 1-1 mapping (like VarChar, Char...)?

Another thing is that to be able to compile, you may also need to modify all the `impl_transport!`. For example, for oracle to arrow data transport in [here](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/oracle_arrow.rs#L41), update into `{ Date[NaiveDateTime]   => Date64[NaiveDateTime]      | conversion auto }` (like timestamp rows). The same applies to files including: [oracle to arrow2](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/oracle_arrow2.rs#L43), [oracle to pandas](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/pandas/transports/oracle.rs#L30).

FerriLuli:
Thank you wangxiaoying for the help, I'm gonna try to do all the modifications and test it, I was'nt cautious enough.

PR #342, opened 2022-09-04
Try to add integrated-auth-gssapi

johnallen3d:
Opening this PR in order to run actions in an attempt to ultimately resolve #334.

Issue #334, opened 2022-08-24
Add `integrated-auth-gssapi` Feature Flag


johnallen3d:
#### Describe your feature request

I'm trying to use ConnectorX with a SQL Server database via Kerberos (trusted connection without username/password). If I'm understanding the [Integrated Authentication documentation on Tiberius](https://github.com/prisma/tiberius#integrated-authentication-trustedconnection-on-nix), we would need to have the [`integrated-auth-gssapi` feature flag](https://github.com/prisma/tiberius#feature-flags) enabled for this to work. Would there be any downside to including that [here](https://github.com/sfu-db/connector-x/blob/c02eb6c5554967115241db171efc5333783913e4/connectorx/Cargo.toml#L50)?

wangxiaoying:
Hi @johnallen3d , thanks for the request.

The feature `integrated-auth-gssapi` relies on some system libraries installed, which requires extra configurations in the CI and release workflows [here](https://github.com/sfu-db/connector-x/tree/main/.github/workflows) on different platforms as well as environments to test this feature.

I tried to build on an ubuntu instance with `libkrb5-dev` and `clang` installed, which succeed. But I encountered the following error when I build the wheel file using `maturin`:
<img width="1571" alt="image" src="https://user-images.githubusercontent.com/5569610/187799651-7ceee9d6-34ca-4df3-bb9a-af1549ef6b50.png">

It is very welcome if you want to help us add this feature. You can follow [this instruction] to build connectorx locally. System dependencies on different platforms can be found in the release script [here](https://github.com/sfu-db/connector-x/blob/main/.github/workflows/release.yml).

johnallen3d:
Thanks for looking into this @wangxiaoying. I will happily take a crack at it!

PR #472, opened 2023-03-16
Libgssapi

kotval:
The Tiberius connector's `AuthMethod::Integrated` means AD on windows, but it means kerberos on *nix. If libgssapi is installed, then you might want to use this auth method. This is a common set up in orgs which mostly use Windows and disallow SA auth. Kerberos is the only way to connect to the MSSQL server from linux. The `#[cfg(windows)]` meant that if you tried to use kerberos from a *nix client to connect to a MSSQL db with tiberius, you'd get a rather cryptic error message. For instance

```import connectorx as cx

con_str = f"mssql://some.db.url:1433/encrypt=true&trusted_execution=true"

query = "SELECT * FROM Table

cx.read_sql(con_str,query)
```

errors out with 

`Login failed for user ''. code=18456`

when run on linux. 

This PR fixes that issue.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install tools
         run: |
           apt-get update
-          apt-get install -y curl postgresql-client build-essential python3-dev python3-pip pkg-config libssl-dev git sqlite3 libsqlite3-dev mysql-client libmysqlclient-dev
+          apt-get install -y curl postgresql-client build-essential python3-dev python3-pip pkg-config libssl-dev git sqlite3 libsqlite3-dev mysql-client libmysqlclient-dev libkrb5-dev libclang-dev
         env:
           DEBIAN_FRONTEND: noninteractive
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Install tools
         run: |
           apt-get update
-          apt-get install -y curl postgresql-client build-essential pkg-config libssl-dev git sqlite3 libsqlite3-dev mysql-client python3 python3-pip libicu66
+          apt-get install -y curl postgresql-client build-essential pkg-config libssl-dev git sqlite3 libsqlite3-dev mysql-client python3 python3-pip libicu66 libkrb5-dev libclang-dev
           pip3 install mssql-cli
           pip3 install cli-helpers==2.2.0
           ln -s /usr/bin/python3 /usr/bin/python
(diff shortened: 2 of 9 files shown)


kotval:
I missed the other PR #342, but even if that one is approved, this issue will remain as the `#[cfg(windows)]` decorator is still not correct for *nix. If that other PR is merged, I can open another PR with only the decorator change. Alternatively, that other one seems stalled for some reason. I can update this PR instead. Please pick one of these options.

wangxiaoying:
Hi @kotval , thanks for the PR.

I think #342 stalled due to the test of building on all platforms. Basically we need to make sure the [release](https://github.com/sfu-db/connector-x/blob/main/.github/workflows/release.yml) workflow can run correctly on linux, mac and windows (not including the final upload task).

It would be great if you can update this PR with modifications from #342 and try to pass both ci and release workflow. Please let me know if you need my help in running the workflow (I'm not sure whether you can run release it on your forked repo locally).

kotval:
Is there a reason in the release build you use yum on what appears to be an ubuntu image? The following is in the release.yaml: https://github.com/kotval/connector-x/blob/main/.github/workflows/release.yml

` runs-on: ubuntu-latest
    container: quay.io/pypa/manylinux2014_x86_64
    strategy:
      matrix:
        python-version: [[38, "3.8"], [39, "3.9"], [310, "3.10"], [311, "3.11"]]
    steps:
      - uses: actions/checkout@v2

      - name: Set python version
        run: |
          if [[ "${{ matrix.python-version[0] }}" == "37" ]]; then
            echo "/opt/python/cp37-cp37m/bin" >> $GITHUB_PATH
          else
            echo "/opt/python/cp${{ matrix.python-version[0] }}-cp${{ matrix.python-version[0] }}/bin" >> $GITHUB_PATH
          fi

      - uses: actions-rs/toolchain@v1
        with:
          toolchain: 1.65.0
          components: rustfmt
          target: aarch64-unknown-linux-gnu
          default: true

      - uses: extractions/setup-just@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

      - uses: Gr1N/setup-poetry@v8

      - name: Install tools
        run: |
          yum install -y mysql-devel postgresql-devel freetds-devel`

I need to add kerberos and clang, but that is different on fedora like systems and ubuntu. I tried adding `krb5-libs` and `llvm-toolset-7.0` which would be what I needed to compile the library on fedora, but ran into issues. https://github.com/kotval/connector-x/actions/runs/4492609578/jobs/7902721615. Can I swap this out for `apt` calls?

It passed on mac and windows.

kotval:
Also, apparently this addressed #334

kotval:
Go ahead and run the workflows that require approval. @maxb2 fixed the workflows to get them to build. If you have any questions about the changes, ask him. Thanks @maxb2 !

maxb2:
An important change to consider is that I had to upgrade from `manylinux2014` to `manylinux_2_28`. It was the only way to get a recent enough version of `clang` so that `libgssapi` could build (other than compiling `clang` from source in `manylinux2014`). For what it's worth, `manylinux2014` will be [EOL on June 30th, 2024](https://github.com/pypa/manylinux#manylinux), so you'll have to upgrade in the near-ish future anyway.

kotval:
There is an issue that I discovered. I'm not sure how to get the integrated-auth-gssapi feature to be available from python. There is something I don't understand about how maturin builds the connectorx crate. I'll try to add a test to the python package that will illustrate the issue.

wangxiaoying:
Thanks @kotval @maxb2 for working this out! I will try to release an alpha version this weekend.

wangxiaoying:
`0.3.2-alpha.3` has been released.

maxb2:
I double checked that the `0.3.2a3` pypi package works as expected. Thanks for the quick turnaround!

PR #525, opened 2023-08-01
Arrow2 and Polars- versions are terribly outdated - updated

AnatolyBuga:
Updated varsions:
```toml
arrow2 = {version = "0.17.4", default-features = false, optional = true}
polars = {version = "0.32.1", optional = true, features=["dtype-u8", "dtype-u16"]}
```


wangxiaoying:
Hi @AnatolyBuga , thanks for the PR!

Looks like the rust test has some linking error. And seems like python test still has some arrow API related issues.

AnatolyBuga:
@wangxiaoying , the branch you linked to has same old versions (arrow2 0.10, polars 0.20 etc)

I have updated connectorx-python/src scripts.

**For me locally both connector-x and connectorx-python build fine** (ie `cargo b --all-features` works).
Rust's linking error is very strange. I could only suggest adding `--release` to `cargo test`

wangxiaoying:
> @wangxiaoying , the branch you linked to has same old versions (arrow2 0.10, polars 0.20 etc)
> 
Ob, right this is a newer version.


> **For me locally both connector-x and connectorx-python build fine** (ie `cargo b --all-features` works). Rust's linking error is very strange. I could only suggest adding `--release` to `cargo test`

I'm not sure using `--release` is a way to fix the issue, and we need to be able to run tests under debug mode.
[Here](https://github.com/sfu-db/connector-x/blob/main/.github/workflows/ci.yml) is the script of the github workflow of our ci, including environment and dependency setup. Maybe it can be useful for you to test locally.

wangxiaoying:
Hi @AnatolyBuga , again thanks for the PR.

I checked the code and fixed the python error [here](https://github.com/sfu-db/connector-x/pull/525/commits/829bcba84bea571cda10feae19b22938c9243305).

As for the rust error, it seems like it is not related to this PR since we saw all other PRs all have this same error: #530

AnatolyBuga:
thank you very much @wangxiaoying for your help and for this great library

PR #529, opened 2023-08-24
fix incorrect nullability for datetime type when writing to arrow

Cabbagec:
It panics when I try to read nullable datetime fields, then I found this incorrect conversion.


wangxiaoying:
Hi @Cabbagec thanks for the bug fix!

PR #533, opened 2023-08-29
build(pypi): Constrain the python version

zen-xu:
I have a py3.7 only project, which is managed by poetry.
But I found that the `connectorx` library does not constrain the python version in the lock file, so It will try to install `connectorx==0.3.2` which has deprecated py3.7 and ultimately fail.
WeChat97df62c932fc01b553cd41f55e8630c5


wangxiaoying:
@zen-xu , thanks for the PR!

PR #542, opened 2023-09-17
Update typesystem.rs, added MYSQL_TYPE_NULL

luca-trigili:


wangxiaoying:
Hi @luca-trigili , thanks for the PR!

Can you also add a test for the null type in our CI? You can refer to this commit: https://github.com/Amar1729/connector-x/commit/3391f71869f25d66e586999bbc6ba8866d99aed6 which adds a new type to postgres. Basically you need to add a new test type `null` to `mysql.sql` and then modify corresponding tests ([here](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/connectorx/tests/test_mysql.py#L276) and [here](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/connectorx/tests/test_mysql.py#L345)) in `test_mysql.py`.

PR #555, opened 2023-11-30
bump arrow to 49 to match latest

AnatolyBuga:


tschm:
This should be covered by a dependabot...

AnatolyBuga:
> This should be covered by a dependabot...

didn't know there is a bot which could do this! If so, it must be a super smart one :D

tschm:
If something is tedious and has to be repeated ===> There is a bot for it...

PR #562, opened 2023-12-13
update to `r2d2-oracle=0.6`

rursprung:
this is the only published crate still depending on v0.5 and is probably the reason for a lot of the downloads of the old version.

note: if you were to enable dependabot updates on connector-x then this wouldn't have to be done manually.


wangxiaoying:
Thanks for the PR @rursprung!

PR #568, opened 2024-01-25
Update SQLite dependencies

aljazerzen:
Versions of SQLite currently used have some reported vulnerabilities (see output of `cargo audit`). This PR updates the rusqlite version to avoid that. 

I've also added a few .gitignore entries for files I need to build this project.


wangxiaoying:
Thanks for the PR @aljazerzen !

PR #573, opened 2024-02-13
Infinite Postgres Bug

AndrewJackson2020:
Fix for a bug with postgres implementation. In postgres many numeric/date data types can have infinite or negative infinite values. Currently when one of these values is selected in connector-x it results in a runtime panic error. This PR fixes this issue for 3 data type (datetime, date, and real).

Resolves Issue 572


AndrewJackson2020:
Issues with current state of fix.

PR #583, opened 2024-03-09
Update dependencies in cargo.toml using a bot

tschm:


tschm:
Related to #555

wangxiaoying:
Thanks @tschm , we can give it a try! (some of our dependencies might not be that stable yet, let's see what will happen first : )

PR #585, opened 2024-03-15
implement Trino source

domnikl:
I'd really like to see Trino being implemented as a source in ConnectorX and thus have begun working on it. It uses prusto as a client and currently supports all basic types. Support for tuple, array, row and uuid isn't implemented yet and there's a bug with time columns not being mapped correctly. Also I want to look into fetching results more efficiently and not everything at once.

What do you think, is it worth continuing work on it? Is there anything else I need to consider eg. regarding mapping the results?

### DIFF
--- a/connectorx-python/connectorx/tests/test_trino.py
+++ b/connectorx-python/connectorx/tests/test_trino.py
@@ -1,0 +1,470 @@
+import os
+
+import pandas as pd
+import pytest
+from pandas.testing import assert_frame_equal
+
+from .. import read_sql
+
+
+@pytest.fixture(scope="module")  # type: ignore
+def mysql_url() -> str:
+    conn = os.environ["MYSQL_URL"]
+    # conn = os.environ["MARIADB_URL"]
+    return conn
+
+
+def test_mysql_without_partition(mysql_url: str) -> None:
+    query = "select * from test_table limit 3"
+    df = read_sql(mysql_url, query)
+    expected = pd.DataFrame(
+        index=range(3),
+        data={
+            "test_int": pd.Series([1, 2, 3], dtype="Int64"),
+            "test_float": pd.Series([1.1, 2.2, 3.3], dtype="float64"),
+            "test_enum": pd.Series(["odd", "even", "odd"], dtype="object"),
+            "test_null": pd.Series([None, None, None], dtype="Int64"),
+        },
+    )
+    assert_frame_equal(df, expected, check_names=True)
+
+
+def test_mysql_with_partition(mysql_url: str) -> None:
+    query = "select * from test_table"
+    df = read_sql(
+        mysql_url,
+        query,
+        partition_on="test_int",
+        partition_range=(0, 10),
+        partition_num=6,
+    )
+    expected = pd.DataFrame(
+        index=range(6),
+        data={
+            "test_int": pd.Series([1, 2, 3, 4, 5, 6], dtype="Int64"),
+            "test_float": pd.Series([1.1, 2.2, 3.3, 4.4, 5.5, 6.6], dtype="float64"),
+            "test_enum": pd.Series(
+                ["odd", "even", "odd", "even", "odd", "even"], dtype="object"
+            ),
+            "test_null": pd.Series([None, None, None, None, None, None], dtype="Int64"),
+        },
+    )
+    df.sort_values(by="test_int", inplace=True, ignore_index=True)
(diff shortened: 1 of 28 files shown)


wangxiaoying:
Thanks @domnikl for the PR. I think it in general looks great and complete!

> Also I want to look into fetching results more efficiently and not everything at once.

Yes, I think this might be an issue in terms of performance. I don't have experience of using prusto before, but it seems like the [`get` and `get_next` APIs](https://docs.rs/prusto/latest/prusto/client/struct.Client.html#method.get) are working for gradually fetching the results. Not sure whether it is easy to switch from `get_all` to these two.

> time columns not being mapped correctly

I'm not sure what exactly the bug is. Currently we are converting the `NaiveTime` type in rust to `String` type in pandas in general (and also I see it in your pr), it is done by the function [here](https://github.com/sfu-db/connector-x/pull/585/files#diff-be5c851d6033675a3bf413eca3c67cc1a38852afbbb50f20ad5511e24ac4dfc0R44). If you want to customize the conversion or mapping it to another type in pandas, it can be done by modifying the mapping in the macro as well as this function. 

It would be great to also submit the corresponding seed database script [here](https://github.com/sfu-db/connector-x/tree/main/scripts) so others can run the python test_trino locally.

connectorx-python/connectorx/tests/test_trino.py
```diff
@@ -0,0 +1,470 @@
...
+
+import pandas as pd
+import pytest
+from pandas.testing import assert_frame_equal
+
+from .. import read_sql
+
+
+@pytest.fixture(scope="module")  # type: ignore
+def mysql_url() -> str:
```
wangxiaoying, line 11: Are these functions should be rename to .*trino.* from mysql?

domnikl:
@wangxiaoying thanks for looking into it! 

> I'm not sure what exactly the bug is

The problem is with reading timestamps with time zone types, prusto does not support it (yet) by the looks of it. Instead it returns `EmptyData` and it appears the result is just empty. But its not that important anyway I think to get starting. I'll create an issue in their project.

> It would be great to also submit the corresponding seed database script

Will do in the next days and also rework fetching from the database with the `get`/`get_next` calls if possible.

domnikl:
@wangxiaoying I implemented the missing partitioning as well as switched to get/get_next for prusto and fixed the tests and added test data and squashed some bugs along the way. Could you have a look at it again please?

connectorx-python/connectorx/tests/test_trino.py
wangxiaoying, line 1: Since we probably are not going to make trino in our CI on github workflow. Can you make all these tests skipped if `TRINO_URL` is not set? Something similar to [oracle](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/connectorx/tests/test_oracle.py#L24) here.

connectorx/src/sources/trino/mod.rs
```diff
@@ -0,0 +1,649 @@
...
+use super::{PartitionParser, Source, SourcePartition};
+
+use anyhow::anyhow;
+
+pub mod errors;
+pub mod typesystem;
+
+#[throws(TrinoSourceError)]
+fn get_total_rows(rt: Arc<Runtime>, client: Arc<Client>, query: &CXQuery<String>) -> usize {
+    rt.block_on(client.get_all::<Row>(query.to_string()))
```
wangxiaoying, line 30: It seems like this will execute the entire query and get all the results, which may be a bottleneck in terms of performance (since the query will be ran twice). Can we make it to using the `SELECT COUNT(*)` query to get the total rows? We have a util function named `count_query`. An example can be found [here](https://github.com/sfu-db/connector-x/blob/52da49ac0d2ddee458cc3279a20f5079b8b754f4/connectorx/src/sources/postgres/mod.rs#L70).

wangxiaoying:
Hi @domnikl , thanks so much for completing the PR!

I've tested the test cases locally and it seems works well. I have two minor comments above but in general I think the code looks good and complete. We probably also need to add a documentation page for trino [here](https://github.com/sfu-db/connector-x/tree/main/docs/databases).

I think after fixing the above minor issues, we can merge it and make it into our next release!

domnikl:
@wangxiaoying thanks for having a look! Added the missing pieces and a bit of documentation for it.

wangxiaoying (approved):
Thanks @domnikl , the update looks good to me! We can merge this PR after the CI is passed : )

Issue #575, opened 2024-03-05
Support SQL ending with a semicolon


theelderbeever:
#### Describe your feature request

Currently, you can't submit queries that terminate with a semicolon. It would be nice to be able to able to include these. Primary reasons being that it is part of the standard already and a lot of syntax highlighters trigger off of the `;` to stop highlighting.

<img width="1739" alt="image" src="https://github.com/sfu-db/connector-x/assets/29295582/a6bed233-73a0-4626-a93b-526255643cf0">

Jordan-M-Young:
Hey @theelderbeever, 

I think this should simple as adding a few well placed query.replace(";","") statements [here](
https://github.com/sfu-db/connector-x/blob/main/connectorx-python/connectorx/__init__.py).

Happy to take a look at this when I have a chance.

theelderbeever:
@Jordan-M-Young Awesome and greatly appreciated!

aimtsou:
Good afternoon,

I have written some code which i think it fixes the problem, I am not sure though if we need to write some extra tests for that because the behavior for the tests should not change.

What do you think @theelderbeever

theelderbeever:
@aimtsou do you have a branch or something that I can look at?

aimtsou:
> @aimtsou do you have a branch or something that I can look at?

https://github.com/aimtsou/connector-x/blob/aimilios_semicolon/connectorx-python/connectorx/__init__.py
Lines 207-212

Jordan-M-Young:
@theelderbeever Just merged this [PR](https://github.com/sfu-db/connector-x/pull/599) Hopefully things are good to go now

PR #599, opened 2024-04-04
Feat/pg semicolon support

Jordan-M-Young:
Implementing support for finishing postgres queries with a semicolon per this [issue](https://github.com/sfu-db/connector-x/issues/575).

Changes include:
     - replace statements to remove semicolon from query in connectorx-python/connectorx/__init__.py (read_sql)
     - Tests to confirm that command terminating semicolons are being removed.

### DIFF
--- a/connectorx-python/connectorx/__init__.py
+++ b/connectorx-python/connectorx/__init__.py
@@ -206,6 +206,7 @@ def read_sql(
     """
     if isinstance(query, list) and len(query) == 1:
         query = query[0]
+        query = query.replace(";","")
 
     if isinstance(conn, dict):
         assert partition_on is None and isinstance(
@@ -214,6 +215,8 @@ def read_sql(
         assert (
             protocol is None
         ), "Federated query does not support specifying protocol for now"
+
+        query = query.replace(";","")
         result = _read_sql2(query, conn)
         df = reconstruct_arrow(result)
         if return_type == "pandas":
@@ -232,6 +235,7 @@ def read_sql(
         return df
 
     if isinstance(query, str):
+        query = query.replace(";","")
         if partition_on is None:
             queries = [query]
             partition_query = None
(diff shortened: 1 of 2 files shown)


Jordan-M-Young:
@wangxiaoying I think this one is ready for review. Let me know if/what changes are needed. Thanks in advance!

wangxiaoying:
Hi @Jordan-M-Young , thanks for the PR!

I think my concern is that ';' may also appears in other parts of the query, such as string matching. For example: `select s from t where s like '%abc;'`. In this case, simply replace ";" with "" will change the semantic of the sql query.

I would suggest to only replace the ";" if it is the last valid character in the sql query. Such as `select s from t;` or `select s from t;[spaces] ` but not `select s from t where s like '%abc;'`

Jordan-M-Young:
@wangxiaoying Thanks for the suggestion. I believe my newest commit should allay that concern. If you wouldn't mind taking another look I'd appreciate it.

wangxiaoying:
Thank you @Jordan-M-Young , I have merged the PR

connectorx-python/connectorx/__init__.py
```diff
@@ -377,3 +385,11 @@ def reconstruct_pandas(df_infos: Dict[str, Any]):
     )
     df = pd.DataFrame(block_manager)
     return df
+
+
+def remove_ending_semicolon(query: str) -> str:
```
surister, line 390: This looks a bit cumbersome. Could be rewritten to:

```python
if query.endswith(';'):
    query = query[:-1]
return query
```

wangxiaoying: Thanks @surister , indeed your suggestion is better. Can you open a PR for this?

surister: You got it.

PR #608, opened 2024-04-17
refactor: A more Pythonic way to set environment variables.

zen-xu:
Signed-off-by: ZhengYu, Xu <zen-xu@outlook.com>


wangxiaoying:
Thanks @zen-xu for the PRs!

PR #621, opened 2024-04-26
Bump datafusion and arrow library to latest version

fredrikbackstrom:
Bump datafusion and arrow version, dependabot PR for arrow failed. Attempting to update both to clear the error

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,5 +8,5 @@ debug = true
 lto = true
 
 [workspace.dependencies]
-arrow = {version = "46", features = ["prettyprint", "ffi"]}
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -8,5 +8,5 @@ debug = true
 lto = true
 
 [workspace.dependencies]
-arrow = {version = "46", features = ["prettyprint", "ffi"]}
+arrow = {version = "51.0.0", features = ["prettyprint", "ffi"]}
```
wangxiaoying, line 11: Can you make it less strict by only keeping the major (and maybe also minor) version like `51` or `51.0` ? Same comment for datafusion.

PR #652, opened 2024-06-20
Change ValueError exception to ModuleNotFoundError

aimtsou:
The problem is analyzed here: https://github.com/sfu-db/connector-x/issues/653


wangxiaoying:
Thanks @aimtsou for the PR!

Issue #510, opened 2023-06-25
Why am I getting [binary data] from MySQL and not original utf8 encoded string?


AnatolyBuga:
#### Why am I getting [binary data] and not utf8 encoded string?

```python

import polars as pl
import connectorx as cx

data = {"a": ["z", "w"], "b": [3, 4]}
df = pl.DataFrame(data)

uri="mysql://{0}:{1}@{2}:{3}/{4}".format(
            "anat" , "XXX", "localhost", 3306, "DBNAME")

df.write_database("test", uri)

df = cx.read_sql(uri, "SELECT * FROM test", partition_num=8, return_type='polars', protocol='text')

print(df)
```

The output is:
```
┌───────────────┬─────┐
│ a             ┆ b   │
│ ---           ┆ --- │
│ binary        ┆ i64 │
╞═══════════════╪═════╡
│ [binary data] ┆ 3   │
│ [binary data] ┆ 4   │
└───────────────┴─────┘
```

column "a" should be a string! What am I doing wrong?

Issue #464, opened 2023-03-07
Mysql text format is interpreted as bytes instead of string


Syndorik:
#### What language are you using?

Python

#### What version are you using?

0.3.1

#### What database are you using?

MySQL

#### What dataframe are you using?

Pandas

#### Can you describe your bug?

When querying a column from a table which has the format type `text`, values are not of type `str` inside the dataframe but of type `bytes`. 

#### What are the steps to reproduce the behavior?

Query a table with text fromat, and check the type of the values (is bytes, should be string)

##### Example query / code

```
df = cx.read_sql(
            "myuri", "SELECT table.text_column FROM table;
        )
```

#### What is the error?

No error, strange behavior

I wanted to note, that this doesn't happen with `pd.read_sql`

mbarki-mohamed:
I'm not sure if I correctly understood your issue, but I tested the code below where I have two columns in my table "varchar_val" contains data of type "varchar" and "text_val" which hold text values.
`import connectorx as cx

query = "SELECT text_val, varchar_val FROM test"
conn = "connectionuri....."
print(cx.read_sql(conn, query)["varchar_val"].dtypes)
print(cx.read_sql(conn, query)["text_val"].dtypes)`

which returns "object" for both, this is the correct behaviour of pandas.

jvovk:
@mbarki-mohamed this is what @Syndorik meant
<img width="87" alt="Screenshot 2023-04-03 at 17 09 15" src="https://user-images.githubusercontent.com/14542319/229551601-e5894c92-8a7e-48a1-8a2b-9542b4d3e8b4.png">
the data inside the column is bytes, not str type

jvovk:
@mbarki-mohamed is it something that is going to be fixed?

Syndorik:
Sorry I totally forgot that issue, the issues deals with what @jvovk describe. The data inside the column are considered as bytes instead of strings

Syndorik:
I'm sorry to reopen the subject, but Is there any work on that planned?

PR #660, opened 2024-07-06
read mysql text column (utf8 charset) as string column

Cabbagec:
Encountered with the same problem as #464 and #510 , where `text` type columns in mysql are interpreted as binary ones naively, which is technically ok, according to the [mysql doc](https://dev.mysql.com/doc/refman/8.4/en/blob.html) that the `text` types are basically just `blob` types, except for additional charset requirements. 

But it's quite an inconvenient for users must do decoding themselves on the final result. So I made and tested this patch in our environment as an easy workaround. Note that we only use UTF8 charset for `text` fields in our databases, so this patch only works for UTF8 text fields. For other encodings like GBK, SHIFT_JIS, BIG5, etc., it won't work, additional decoding is still needed.

If anyone interested, I recommend checking out this crate [`encoding_rs`](https://docs.rs/encoding_rs/latest/encoding_rs/), then doing decoding steps according to the mysql field charset info.

### DIFF
--- a/connectorx/src/sources/mysql/mod.rs
+++ b/connectorx/src/sources/mysql/mod.rs
@@ -17,7 +17,14 @@ use fehler::{throw, throws};
 use log::{debug, warn};
 use r2d2::{Pool, PooledConnection};
 use r2d2_mysql::{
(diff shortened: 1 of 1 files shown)


wangxiaoying:
Thanks @Cabbagec for the PR and testing!

Issue #491, opened 2023-04-20
DataType BIT for mysql is missing


x14119641:
#### I found this bug using Python (Polars) and the latest connectorx to connect to MariaDB.

I would appreciate if someone can add it. 
I tried by myself but I encountered loads of problems to run this project with rust in windows.

The DataType is: **MYSQL_TYPE_BIT** 

#### What is the error?

```
pyo3_runtime.PanicException: not implemented: MYSQL_TYPE_BIT
```

**Thanks in advance!!**

MidasLamb:
I'm also encountering this issue. I'm trying to make a PR to fix this but some of the main parts are macro's, which make it quite difficult to figure out what has to be done. If someone could give a high level overview, that would be an immense help

x14119641:
You have some info [here](https://github.com/sfu-db/connector-x/blob/main/CONTRIBUTING.md) but most of the links are broken. I rembember to find an example when I opened this issue, similar [to this one](https://github.com/sfu-db/connector-x/tree/main/connectorx/src/sources) but i belive more things are need to be changed in order to make it work.

I might give another try to fix this during the weekend, my problem are incompatibilities with different versions when i try to load the project using "cargo" and "just".

MidasLamb:
I think I have something working, but it maps it to a `Vec<u8>`, since you can define i.e. `BIT(64)` .
If you use a `BIT(1)` for a boolean value, that might be annoying to work around, but I don't think I can make it be a boolean for `BIT(1)` and an array for all the other values.

dawsongzhao0523:
I have also encountered the same issue. What is the current progress of it?

surister:
It'd be helpful to provide a piece of code to reproduce the error

Issue #359, opened 2022-09-22
PanicException: not implemented: MYSQL_TYPE_BIT


Yanhuanjin:
#### What language are you using?

**Python**

#### What version are you using?

0.3.0

#### What database are you using?

MySQL

#### What dataframe are you using?

Pandas

#### Can you describe your bug?

When the queried database field was bit, an error was reported.

##### Example query / code
I create a table and the data looks like bellow.
![image](https://user-images.githubusercontent.com/25056602/191644869-0b479046-22e3-4de3-88fb-49ee9c9b20eb.png)
![image](https://user-images.githubusercontent.com/25056602/191644886-2129fbcc-506b-4255-8c5c-002a8acc6cad.png)

When I queried this table, error occuried.
```
df = cx.read_sql(conn, "SELECT * FROM test;")
```

#### What is the error?

PanicException                            Traceback (most recent call last)
<ipython-input-46-55e1bdb14253> in <module>
----> 1 df = cx.read_sql(conn, sql_list)

C:\ProgramData\Anaconda3\lib\site-packages\connectorx\__init__.py in read_sql(conn, query, return_type, protocol, partition_on, partition_range, partition_num, index_col)
    222             raise ValueError("You need to install pandas first")
    223 
--> 224         result = _read_sql(
    225             conn,
    226             "pandas",

PanicException: not implemented: MYSQL_TYPE_BIT

wangxiaoying:
BIT type is not supported in mysql yet: https://sfu-db.github.io/connector-x/databases/mysql.html

florianlaws:
I would also appreciate this being suported.

PR #661, opened 2024-07-06
read mysql bit type as i64

Cabbagec:
Currently, mysql column type `MYSQL_TYPE_BIT` is not supported, panicking the process if present in the table. And this issue is also stated in #359 and #491 .

I think it's appropriate to convert the bit type as longlong i64. According to the [mysql doc](https://dev.mysql.com/doc/refman/8.4/en/bit-type.html) page, bit type columns can store `M`-bit values, with `M` ranging from 1 to 64.

### DIFF
--- a/connectorx-python/connectorx/tests/test_mysql.py
+++ b/connectorx-python/connectorx/tests/test_mysql.py
@@ -337,6 +337,9 @@ def test_mysql_types_binary(mysql_url: str) -> None:
             "test_mediumtext": pd.Series(
                 [None, b"", b"medium text!!!!"], dtype="object"
             ),
+            "test_bit": pd.Series(
+                [b'\x17', b'\x18', None]
+            ),
         },
     )
     assert_frame_equal(df, expected, check_names=True)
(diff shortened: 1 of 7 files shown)


wangxiaoying:
Thanks @Cabbagec for the PR! Since it's a new type, can you also add it to our unit test? You can add a new column and insert some values in the mysql table defined [here](https://github.com/sfu-db/connector-x/blob/main/scripts/mysql.sql#L33), and modify the unit corresponding tests [here](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/connectorx/tests/test_mysql.py).

Cabbagec:
> Thanks @Cabbagec for the PR! Since it's a new type, can you also add it to our unit test? You can add a new column and insert some values in the mysql table defined [here](https://github.com/sfu-db/connector-x/blob/main/scripts/mysql.sql#L33), and modify the unit corresponding tests [here](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/connectorx/tests/test_mysql.py).

Sure, will do later.

PR #666, opened 2024-07-27
Add support for pgpass

surister:


surister:
@wangxiaoying Adding a bunch of new code to support specific postgres functionality into the `__init__` feels bloat, how about we add new code in a new file? maybe something like `postgres/pgpass.py`?

 `__init__.py`
 `postgres/pgpass.py`

connectorx-python/connectorx/__init__.py
```diff
@@ -247,6 +247,66 @@ def read_sql(
 ) -> pl.DataFrame: ...
 
 
+def get_passfile_content(path: pathlib.Path) -> dict:
+    # host:port:db_name:user_name:password
+    with open(path) as f:
```
aimtsou, line 252: How about we do something different? the content we need from pgpass is the password only the restof the elements are used for retrieving it. So:

```python
            for line in f:
                entry = read_pgpass_line(line)
                if entry:
                    if ((entry['host'] == host or entry['host'] == '*') and
                        (entry['port'] == port or entry['port'] == '*') and
                        (entry['database'] == database or entry['database'] == '*') and
                        (entry['user'] == user or entry['user'] == '*')):
                        return entry['password']
```

connectorx-python/connectorx/__init__.py
```diff
@@ -247,6 +247,66 @@ def read_sql(
...
+        contents = f.read().split(':')
+        if len(contents) != 5:
+            raise Exception('Pgpass content should follow: host:port:db_name:user_name:password')
+    return dict(zip(['hostname', 'port', 'path', 'username', 'password'], contents))
+
+
+def get_pgpass(conn) -> dict:
+    # check param or (PGPASS env or DEFAULT)
+    # DEFAULT = linux or windows
+    passfile = '%APPDATA%\postgresql\pgpass.conf' if sys.platform == 'windows' else os.path.expanduser(
```
aimtsou, line 262: Instead of sys.platform maybe we can use os.name == 'nt'

connectorx-python/connectorx/__init__.py
```diff
@@ -247,6 +247,66 @@ def read_sql(
...
+
+        # test if there is no '&'
+        for param in parsed_conn.params.split('&'):
+            k, v = param.split('=')
+            if k == 'passfile':
+                passfile = k
+
+    passfile_path = pathlib.Path(passfile)
+
+    if sys.platform != 'windows':
```
aimtsou, line 275: Same here: os.name != 'nt'

connectorx-python/connectorx/__init__.py
```diff
@@ -247,6 +247,66 @@ def read_sql(
...
+                f'pgpass file does not have safe permissions (0600), it currently has "0{st_mode}" you can fix this buy running: $ chmod 0600 PASSFILE_PATH')
+
+    return get_passfile_content(passfile)
+
+
+def replace_conn_content(conn: str, contents: dict) -> str:
+    # We rewrite the netloc from scratch here e.g.
+    # netloc = contents['username'] or o.username + ':' + contents['password'] or o.password + ...
+    parsed_conn = urllib.parse.urlparse(conn)
+    return str(parsed_conn._replace(**contents))
```
aimtsou, line 298: We can do the following:  
    database = parsed_conn.path[1:] or os.environ.get('PGDATABASE', os.getenv('USER'))
    host = parsed_conn.hostname or os.environ.get('PGHOST', 'localhost')
    port = parsed_conn.port or os.environ.get('PGPORT', '5432')
    user = parsed_conn.username or os.environ.get('PGUSER', os.getenv('USER'))
    password = parsed_conn.password

We parse 2 times the URI so maybe we need to rethink that maybe we call pgpass from inside reconstruct? That would allow something like that also:

query_params = parse_qs(parsed_con.query)

    passfile = query_params.get('passfile', [None])[0] or os.environ.get('PGPASSFILE')
    default_pgpass_path = os.path.expanduser('~/.pgpass') if os.name != 'nt' else os.path.join(os.getenv('APPDATA', ''), 'postgresql', 'pgpass.conf')

    # Determine the password precedence
    if not password:
        if 'PGPASSWORD' in os.environ:
            password = os.environ['PGPASSWORD']
        else:
            pgpass_path = passfile or default_pgpass_path
            password = get_passfile_content(pgpass_path, host or '*', str(port or '*'), database or '*', user or '*')

And with all these information we can get the new connection string.

connectorx-python/connectorx/__init__.py
```diff
@@ -247,6 +247,66 @@ def read_sql(
...
+
+
+def run_per_database(conn):
+    # Todo rename to something better.
+    if 'postgresql' in conn:
```
aimtsou, line 303: Or postgres is also valid: https://www.postgresql.org/docs/current/libpq-connect.html

wangxiaoying:
> @wangxiaoying Adding a bunch of new code to support specific postgres functionality into the `__init__` feels bloat, how about we add new code in a new file? maybe something like `postgres/pgpass.py`?
> 
> `__init__.py` `postgres/pgpass.py`

Thanks for the PR! I think moving the logic into an independent file would be great!

wangxiaoying:
I just pushed a [commit](https://github.com/sfu-db/connector-x/commit/d22646456f73202e6fc1117f8396979d5014dd41) to fix the error that cause the fail of the CI. Please feel free to rebase on that one.

connectorx-python/connectorx/__init__.py
surister: is it better?

connectorx-python/connectorx/__init__.py
surister: Then we ought to change it to `postgres`

surister:
Sorry I don't have time to finish this

PR #693, opened 2024-10-02
Update Federated Rewriter dependencies

abecker4:
There are a few vulnerabilities within the Federated Rewriter Jar file for the python connector-x. This PR Has a jar with updated dependencies. The only issue is the jar file was packaged without tests, so it would probably be better to update in via the Federated Rewriter repository. I just was unable to find it :D

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -46,7 +46,7 @@ jobs:
         run: touch ./docs/_build/html/.nojekyll
 
       - name: Deploy Docs 🚀
-        uses: JamesIves/github-pages-deploy-action@v4.6.4
+        uses: JamesIves/github-pages-deploy-action@v4.6.8
         with:
           branch: gh-pages # The branch the action should deploy to.
           folder: ./docs/_build/html # The folder the action should deploy.


wangxiaoying:
Hi @abecker4 , thanks for the PR!

We are considering removing this jar from this repo, and open another repo for federation related logic. 

For now I think if the jar can pass the CI, we can merge it : )

abecker4:
Hello @wangxiaoying, happy to assist!

It looks like the only two tests that are failing are the `test_fed_spj` and `test_fed_spja`. I assume these are due to the dependency updates as they look to be directly related.  

I see that the python tests passed in CI last week, so I took a fresh clone of the connector-x repo and rebuilt without my changes. Unfortunately, I ran into the same Java errors locally that we are seeing in CI with my changes.  I tried to troubleshoot the issues via python shell but ran into the same issue there as well.

As I am not a Java expert, any guidance on what steps I can take to further troubleshoot these two failing tests would be greatly appreciated.

wangxiaoying:
Hi @abecker4 , I have [moved the rewriter jar file](https://github.com/sfu-db/connector-x/blob/main/Federation.md) to another repo: https://github.com/sfu-db/accio and disabled all federation related tests in the CI of connectorx. So the current main branch and from the next release, there will be no jar file included in the wheel. I guess we can close this issue for now. But feel free to reopen this.

Issue #509, opened 2023-06-23
Cannot connect to MSSQL DB -> RuntimeError: Timed out in bb8


mavestergaard:
#### What language are you using?

**Python**.

#### What version are you using?

Python 3.10
connectorx==0.3.1

#### What database are you using?

MSSQL

#### What dataframe are you using?

Polars

#### Can you describe your bug?

I am trying to connect to our Microsoft SQL Database. It works with telnet, SQL Server Management Studio and pyodbc. When I connect with connectorx i get error:
RuntimeError: Timed out in bb8

#### What are the steps to reproduce the behavior?

Here are the strings I have tested:
from urllib import parse
import connectorx as cx

# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?driver=ODBC+Driver+18+for+SQL+Server"
# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?driver=ODBC+Driver+17+for+SQL+Server"
# return f"mssql+pyodbc://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?driver=ODBC+Driver+18+for+SQL+Server"
# return f"mssql+pyodbc://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?driver=ODBC+Driver+17+for+SQL+Server"
# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}"
# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?driver=SQL+Server"
# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?trusted_connection=true&encrypt=true"
# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?trusted_connection=false&encrypt=true"
# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?trusted_connection=false&encrypt=false"
# return f"mssql://{username}:{parse.quote_plus(pw)}@{servername}/{dbname}?trusted_connection=true&encrypt=false"

##### Database setup if the error only happens on specific data or data type

Cannot connect to the db, so query doesnt matter

##### Example query / code

Same as above
cx.read_sql(conn = con_str, query = "SELECT 1")

#### What is the error?

RuntimeError: Timed out in bb8

Really hope you can help - thank you very much :D

CodeNinja2C4U:
[...]

PR #698, opened 2024-10-17
feat: Added Params for CA Certs in MSSQL Config

pangjunrong:
When establishing an SSL connection to MSSQL currently, bb8 times out due to a mismatch in CA certificates since underlying default behavior from _tiberius_ is to validate server certificate against system-truststore. 

This change enables 2 new parameters in the connection string for MSSQL to address custom CA certificate use cases ([description extracted from _tiberius_ config documentation](https://docs.rs/tiberius/latest/tiberius/struct.Config.html)):

- **trust_server_certificate**: If set, the server certificate will not be validated and it is accepted as-is
- **trust_server_certificate_ca**: If set, the server certificate will be validated against the given CA certificate in in addition to the system-truststore. Useful when using self-signed certificates on the server without having to disable the trust-chain.

This will resolve issue #509 and is a requested feature at the end of the thread.

### DIFF
--- a/connectorx/src/sources/mssql/mod.rs
+++ b/connectorx/src/sources/mssql/mod.rs
@@ -84,6 +84,16 @@ pub fn mssql_config(url: &Url) -> Config {
         decode(url.password().unwrap_or(""))?.to_owned(),
     ));
 
+    match params.get("trust_server_certificate") {
+        Some(v) if v.to_lowercase() == "true" => config.trust_cert(),
+        _ => {},
+    };
+
+    match params.get("trust_server_certificate_ca") {
+        Some(v) => config.trust_cert_ca(v),
+        _ => {},
+    };
+
     match params.get("encrypt") {
         Some(v) if v.to_lowercase() == "true" => config.encryption(EncryptionLevel::Required),
         _ => config.encryption(EncryptionLevel::NotSupported),


wangxiaoying:
Hi @pangjunrong , thanks for the PR. It looks good to me.

Can you fix the [formatting problem](https://github.com/sfu-db/connector-x/actions/runs/11383850412/job/31711777889?pr=698) of the code? You can do it by running `cargo fmt`.

Also, can you add this new parameter to the documentation [here](https://sfu-db.github.io/connector-x/databases/mssql.html) as well? You can do this by editing [this markdown file](https://github.com/sfu-db/connector-x/blob/main/docs/databases/mssql.md).

pangjunrong:
Hey @wangxiaoying, thanks for the quick reply! I've fixed the formatting and added the new parameters into the documentation.

So I updated tiberius & bb8-tiberius to 0.7.3 and 0.8.0 respectively as those versions contain the first instances of _ca function but I noticed that anything above tiberius 0.5.16 no longer contained the QueryResult method (unfortunately deprecated) which is used across the MSSQL implementation. I've tried my hand at replacing it with the new QueryStream method; hoping you can help out on this if you see anything amiss!

wangxiaoying:
Thank you @pangjunrong !

Issue #202, opened 2021-12-20
Oracle differences between SQLAlchemy and TNS integration.


mdeakyne:
The below code connects to an Oracle Database - using tnsnames.ora, in TNS_ADMIN / ORACLE_HOME.

```python
from os import getenv
from sqlalchemy import create_engine
import pandas as pd

con_string = f'oracle+cx_oracle://{getenv("user")}:{getenv("db_pwd")}@DEMISDW'
conn = create_engine(con_string)

test = pd.read_sql(query, conn)
print(test) #works!
```

Using connectorX 

```python
import connectorx as cx
test = cx.read_sql(f"oracle://{getenv('user')}:{getenv('db_pwd')}@DEMISDW", query = query)
#RuntimeError: timed out waiting for connection: OCI Error: ORA-12154: TNS:could not resolve the connect identifier specified
```

I've tried reconfiguring the host to various formats, and urls - even directly substituting the full string from the TNS File:
```
DEMISDW = (DESCRIPTION=(ADDRESS_LIST=(ADDRESS=(PROTOCOL=TCP)(HOST=<host>)(PORT=1521)))(CONNECT_DATA=(SERVICE_NAME=<service>)))
```

I've also run SQL to try to get other values to try:

```
SELECT * FROM SYS.GLOBAL_NAME; --Gets Service Name
SELECT SYS_CONTEXT('USERENV', 'SERVER_HOST') FROM DUAL; --Gets host
```

Can you provide some more insight on how the TNS works? Is there a location the library looks to? Or an environment variable it pulls from?

Thanks for any guidance you can provide.

wseaton:
@mdeakyne, the underlying library that connects to oracle is `oracle-rs`, here are the docs, I'm wondering if the feature you are asking about is: https://docs.rs/oracle/0.5.3/oracle/struct.Connector.html#method.external_auth

If so, some code could probably be added to the connection builder to dynamically enable this on the rust side here, it would just require passing in a custom oracle connection into the r2d2-oracle connection pool: 
 - https://github.com/wseaton/connector-x/blob/main/connectorx/src/sources/oracle/mod.rs
 
 ```rust
     /// Initialise the connection manager with the data needed to create new connections using `oracle::Connector`.
    /// This allows setting additional connection data.
    ///
    /// If a connection can be established only with a username, password and connect string, use `new` instead.
    ///
    /// # Example
    /// ```
    /// # use r2d2_oracle::OracleConnectionManager;
[...]

Issue #484, opened 2023-04-09
Oracle connection string with TNS


yxiao315:
By using `conn = 'oracle://username:password@server:port/database'` I can connect to Oracle , but How I can use TNS entry name in connectorx? for example, in cx_Oracle, it can be done like this: `conn = cx_Oracle.connect(dsn="the_tns_entry_name")`. Is there similar way to do it in connectorx? Since in my tns entry, it doesn't need user id and password, seems like Windows Authentication.  

Or how to use connectorx connect to Oracle by Window authentication?

mondjef:
also wondering how to use TNS.... I have been able to connect using cx_Oracle with TNS but have been unsuccessful getting connectorX to connect to same oracle db.  I have tried various combinations of connection string based on info in the tnsnames.ora file

Issue #420, opened 2022-12-08
Issue Connecting to Oracle


ciollid2:
Hello,

I am having an issue with the connection string required by connectorx.readsql().
I am using the template suggested by the documentation: 'oracle://username:password@server:port/database' 
However, I get the error:  ORA-12154: TNS:could not resolve the connect identifier specified

I can connect to this same Oracle DB through SQLAlchemy by using the following connection string:
oracle+cx_oracle://{user}:{password}@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST={host})(PORT={port}))(CONNECT_DATA=(SERVICE_NAME={service_name}))

The portion of the string after the '@' is generated by the cx_Oracle.make_dsn function. My username & password contain only alphanumeric characters.

Any help would be greatly appreciated. This package seems to have a ton of potential and I'm excited to try it out.

wangxiaoying:
Hi @ciollid2 , is your servername has special characters? Have you tried to urlencode your servername? Also are you using the newest version?

ciollid2:
Thanks for the quick reply
I am on connectorx 0.3.1 and Python 3.10.5.
There are no special characters in any information I am using in the string. Just to be sure I did try URL encoding everything anyway and it did not work.

ciollid2:
Hello,
Sorry for the late reply. I have gone through each of the Oracle issues posted here and none of the solutions there worked.

My SQLAlchemy connection to our Oracle DB does not work formatted without the make_dsn information. Is it possible the (PROTOCOL=TCP) is required and not being passed? I am not too knowledgeable when it comes to networking.

Thanks

wangxiaoying:
> Is it possible the (PROTOCOL=TCP) is required and not being passed?

I think `(PROTOCOL=TCP)` is is not the issue since others mentioned the description string also contain the same `(PROTOCOL=TCP)`.

Can you try to use command line like `sqlplus` to test (in the same env with your code). Basically try to see whether this works:
```bash
sqlplus ${ORACLE_USER}/${ORACLE_PASSWORD}@${ORACLE_HOST}:${ORACLE_PORT}/${ORACLE_DB}
```

ciollid2:
The sqlplus template you provided did not work for me, however I could connect with the following format:
`sqlplus user/password@SID`

Thanks

wangxiaoying:
> The sqlplus template you provided did not work for me

[...]

PR #700, opened 2024-10-25
feat: Added Option for Oracle Connector to use Alias w/ Params

pangjunrong:
The use of aliases defined in tnsnames.ora is a common connection method found in Oracle, which also enables the specification of communication protocols like TCP/TCPS  (i.e. SSL-enabled) among other configurations through a DSN connection string.

This change adds parameter handling for the connection string defined for an Oracle DB connection. Specifically, the alias boolean parameter when set to true, parses the alias defined in the host directly to the oracle::Connector::new method to initialize the connector object.

For a connection using aliases, the connection string will look as such:
`oracle://<username>:<password>@<alias>?alias=true`
compared to the direct connection string method specifying host & port:
`oracle://<username>:<password>@<host>:<port>/<database>`

This will resolve issues #202, #420 & #484 — it is a requested feature for DSN connection strings in TNS aliases to be used.

### DIFF
--- a/connectorx/src/sources/oracle/mod.rs
+++ b/connectorx/src/sources/oracle/mod.rs
@@ -57,13 +57,13 @@ pub fn connect_oracle(conn: &Url) -> Connector {
     let user = decode(conn.username())?.into_owned();
     let password = decode(conn.password().unwrap_or(""))?.into_owned();
     let host = decode(conn.host_str().unwrap_or("localhost"))?.into_owned();
-    let port = conn.port().unwrap_or(1521);
 
     let params: HashMap<String, String> = conn.query_pairs().into_owned().collect();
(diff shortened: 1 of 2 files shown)


pangjunrong:
Hey @wangxiaoying if this change is good to go, any chance we can get a minor release onto pypi? It'll be really helpful to get the MSSQL SSL support & Oracle alias connection in for downstream uses to test it out 😄

wangxiaoying:
Thank you @pangjunrong for the PR! It looks good to me.

wangxiaoying:
Hi @pangjunrong , I just released an alpha version for the new features. You can install by `pip install connectorx==0.3.4a3`. Please feel free to give it a try!

Issue #320, opened 2022-07-28
Can't connect to MSSQL


SergKS8:
The problem is that
when I'm trying to read data from MSSQL like this:
conn = f'mssql://{login}:{password}@{host}/{database}'
cx.read_sql(conn, sql_aims)

I'm getting the error
RuntimeError: parse error: invalid port number.
But default port = 1433 is a valid port.

But if I changed the password in conn string to combination without symbol '#' I get another arror -
RuntimeError: Timed out in bb8.
I think that thing is that my password contain specific symbol # and this is the cфuse of the error.
How can I read data from MSSQL with connectorx in such a case?

The SOURSE - MS SQL Server with SQL Server Authentication.

wangxiaoying:
Follow up in https://github.com/sfu-db/connector-x/discussions/319

PR #708, opened 2024-11-18
Modifies EncryptionLevel to Off instead of NotSupported

### LINKED ISSUE #707: ConnectorX MsSQL encryptionlevel TimedOut error

Rust

0.3.3 as available in [crates.io](https://crates.io/crates/connectorx), but in addition I checked the [change log](https://github.com/sfu-db/connector-x/compare/v0.3.3...v0.4.0) of `0.4.0` which does not change the relevant lines of code to mitigate the bug. However, I didn't test it yet with `0.4.0`.

MS SQL

Arrow

Using the connectorx library with the Arrow DataFrames in Rust, an issue arises when the encryption level is set to NotSupported. This configuration causes an error during execution. Modifying the encryption level to Off resolves the problem.

If possible, please include a **minimal simple** example including:
```rust 
let destination = get_arrow(&source_conn, None, &[queries.query.clone()]).expect("run failed");
```
It is based on the examples mentioned [here](https://docs.rs/connectorx/latest/connectorx/). Where I use the default connection string. After a lot of debugging, the reason of the bug can be found in this function: [src/sources/mssql/mod.rs](https://github.com/sfu-db/connector-x/blob/4fc8b2511f902e225d6a350a69f89f866c22a391/connectorx/src/sources/mssql/mod.rs#L154).

The suggested fix is:
```
+++ b/connectorx/src/sources/mssql/mod.rs
@@ -96,7 +96,7 @@ pub fn mssql_config(url: &Url) -> Config {

     match params.get("encrypt") {
         Some(v) if v.to_lowercase() == "true" => config.encryption(EncryptionLevel::Required),
-        _ => config.encryption(EncryptionLevel::NotSupported),
+        _ => config.encryption(EncryptionLevel::Off),
     };
```
where we change the `EncryptionLevel::NotSupported` into `EncryptionLevel::Off`.  As a result, we do not get the bug:
```
run failed: MsSQLArrowTransportError(Source(MsSQLRuntimeError(TimedOut)))
```

Not applicable, it holds for any table schema and data because it can't connect to the database at all.

##### Example query / code
See example in the `replicate bug` section.

#### What is the error?
We continuously get an `TimedOut` error:
```
run failed: MsSQLArrowTransportError(Source(MsSQLRuntimeError(TimedOut)))
```

boersmamarcel:
Resolves #707

### DIFF
--- a/connectorx/src/sources/mssql/mod.rs
+++ b/connectorx/src/sources/mssql/mod.rs
@@ -96,7 +96,7 @@ pub fn mssql_config(url: &Url) -> Config {
 
     match params.get("encrypt") {
         Some(v) if v.to_lowercase() == "true" => config.encryption(EncryptionLevel::Required),
-        _ => config.encryption(EncryptionLevel::NotSupported),
+        _ => config.encryption(EncryptionLevel::Off),
     };
 
     match params.get("appname") {


boersmamarcel:
@pangjunrong could you help out with this merge? If there are additional steps I need to take then please let me know. Or could you refer me to somebody else to help get this merge approved?

pangjunrong:
So I spent some time understanding the underlying mechanisms for the different encryption options from [[MS-SSTDS]: Tabular Data Stream Protocol Version 4.2](https://learn.microsoft.com/en-us/openspecs/sql_server_protocols/ms-sstds/dab36a48-6c13-44c7-954a-0f5c8623590d) (specifically 2.2.6.4) and came across several possible issues with using `EncryptionLevel::Off`, such as [tiberius #320](https://github.com/prisma/tiberius/issues/320).

I think the most comprehensive way forward would be to account for all possible DB setups — in your case, the expected connection might not be encrypted but `EncryptionLevel::Off` is necessary because a TLS handshake is still expected during `PRELOGIN`. For someone else, their setup might not support TLS at all and having `EncryptionLevel::Off` might lead to a timeout as the client attempts a secured handshake.

May I propose that when `encrypt=false`, we do `EncryptionLevel::Off` and otherwise we default to `EncryptionLevel::Unsupported`?

For the merge to be approved, you will need @wangxiaoying

wangxiaoying:
Thanks @boersmamarcel for the PR!

I agree with @pangjunrong that we might want to have both `EncryptionLevel::Off` and `EncryptionLevel::NotSupported` supported.

PR #716, opened 2024-12-09
#643 allow to skip verification of TLS certs for Trino connection

### LINKED ISSUE #643: How to use self signed cert with trino connection
Trino do not accept basic authentication using http.

Using https it requires tls/ssl cert, the trino-python library circumvent this by using verify=false.

Is there a way to pass the verify=false as a parameter in the connection string or any other way to connect as the current documentation propose trino+https with the username and password in the connection string; this couldn't work.

Thanks a lot guys!

domnikl:
Resolves #643


tooptoop4:
@wangxiaoying do u know when this can be released?

wangxiaoying:
Hi @tooptoop4 @domnikl , I have released a new alpha version for the PR. Please feel free to check out by `pip install connectorx==0.4.1a1`

Issue #438, opened 2023-01-27
Support for Trino


romanzdk:
Please add support for Trino SQL engine

antonysouthworth-halter:
+1 on this, I feel like it should not be too difficult given there is already the official DBAPI wrapper https://github.com/trinodb/trino-python-client

zeddit:
+1, but it seems trino is using http as interface, thus won't be accelerated too much.

wseaton:
Here's an example Trino Rust client implementation: https://github.com/joshthoward/trino-rust-client, it's basically just a wrapper around a few REST API calls. There are also very tricky things like supporting OAuth and automatically de-serializing the data into an Arrow table format that are absent. 

Benchmarking that (as the best case) against the python bindings might show if there is a room for speedup, but I doubt much is to be gained.

domnikl:
Trino support was added in 0.3.3

JakkuSakura:
Arrow table format is not supported yet

JakkuSakura:
I continued work since the last PR. It now builds with trino arrow support, but I can't use the wheel as it says
ImportError: dlopen(/opt/homebrew/lib/python3.13/site-packages/connectorx/connectorx.cpython-313-darwin.so, 0x0002): weak-def symbol not found '_PyBaseObject_Type'

It's not something I'm familiar with. Maybe it's better to let the maintainers test and build

PR #722, opened 2025-01-06
Trino arrow + arrow2 support

JakkuSakura:
I see the trino arrow + arrow2 transports were added, but still can't use them. I tried to add them here.
closes https://github.com/sfu-db/connector-x/issues/438


wangxiaoying:
Thank you @JakkuSakura for the PR!

Issue #728, opened 2025-01-18
Add support to run pre-execution queries such as SET statements for runtime configurations


jsjasonseba:
#### Describe your feature request

Please describe the behavior you want and the motivation.

Related to:  https://github.com/sfu-db/connector-x/issues/645 https://github.com/sfu-db/connector-x/issues/656 https://github.com/sfu-db/connector-x/issues/512 

Several use cases require issuing SET statement, such as setting transaction isolation levels, timeouts, etc. Connectorx can support this use case by adding a new parameter `pre_execution_querries` on `cx.read_sql` where user can submit SET statements and it is run before the SELECT statement. Since SET statements are typically applied per session (connection), each connection created for each partition must execute the pre-execution queries after connection is created.

I’m currently developing this functionality and planning to submit a PR. But I'm not certain if this aligns with the project’s direction. Any feedback or discussion would be appreciated.

mcrumiller:
Wasn't this issue closed by #729?

jsjasonseba:
> Wasn't this issue closed by [#729](https://github.com/sfu-db/connector-x/pull/729)?

Yes we can close this for now. Altough, we haven't supported this for all data sources. Maybe we can open a separate issue.

Issue #656, opened 2024-06-25
How to set database connection timeout for Oracle database?


fritzwijaya:
I am using the connectorx to export large data from Oracle database. But, facing connection timeout issue.
How to set the connection timeout or increase the connection timeout on the connection url? Regards

ssassi:
We’re experiencing the same connection timeout issue. Any updates or suggested solutions would be appreciated.

Issue #645, opened 2024-06-11
Set TRANSACTION ISOLATION LEVEL ?


kmatt:
Queries appear to default to isolation level READ COMMITTED.

If a different level was desired, such as `SET TRANSACTION LEVEL READ UNCOMMITTED` was desired, can this be set in Connector-X?

Adding them as an initial separate statement to the SQL query results in an error:

`RuntimeError: MsSQL failed to get the columns of query: SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED`

melissa-ellevation:
hi there - I am experiencing the same issue. I imagine if there were any updates someone would have posted them by now...

jsjasonseba:
Hi @kmatt @melissa-ellevation , you can now submit SET statements on `pre_execution_query` parameter on connectorx v0.4.2 for mysql and postgres source. (https://sfu-db.github.io/connector-x/api.html)

kmatt:
Is this feature only supported on Postgres and MySQL?

Issue #512, opened 2023-07-03
How to set timeout for MySQL


akbarnotopb:
#### Try Discussion?

Halo, as the title stated, how to set custom mysql timeout? Im having error when my source database is slow . The error said connection timeout, yet i can't find any parameter to set it. Do i set it inside the connection uri ? or what?

Thanks in advance.

evbo:
Use the JDBC URI query parameters supported by each RDBMS. For instance:

```python
f"mysql://{u}:{p}@{h}:{p}?connect-timeout=500"
```

more here:
https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html

jsjasonseba:
Hi @akbarnotopb  @evbo , you can now submit SET statements in pre_execution_query parameter in connectorx v0.4.2 for mysql and postgres source. You can use SET statements to set wait_timeout or connect_timeout.

```
pre_execution_query = [
  "SET SESSION connect_timeout = 2151",
  "SET SESSION wait_timeout = 2252",
]
df = read_sql(mysql_url, query, pre_execution_query=pre_execution_query)
```

PR #729, opened 2025-01-21
Add pre_execution_queries parameter to run setup queries before main query on Postgres and MySQL source

jsjasonseba:
This PR attempts to solve: https://github.com/sfu-db/connector-x/issues/728
Also related issues: https://github.com/sfu-db/connector-x/issues/645 https://github.com/sfu-db/connector-x/discussions/471 https://github.com/sfu-db/connector-x/issues/656 https://github.com/sfu-db/connector-x/issues/512

This PR adds pre_execution_queries parameter to run setup queries before main query. SET statements are typically applied on session/connection scope, so the pre-execution queries must be applied to each connection on each partition. This is implemented by running the pre-execution queries using a connection before assigning the connection to SourcePartition. This will allow all queries run by each connection on each partition to have the necessary settings.

Currently implemented in Postgres and MySQL sources.

I am quite new with rust and I noticed this change is quite breaking to the rust API since rust does not support default args. Please let me know if there is a better way to do this.


jsjasonseba:
Hi @wangxiaoying, I have fixed the tests and applied formatting. Apologies for missing them earlier. Could you please rerun the workflow?

wangxiaoying:
Thank you @jsjasonseba for the PR! I will take a deeper look at it by this weekend.

connectorx/src/sources/bigquery/mod.rs
```diff
@@ -107,6 +107,8 @@ where
         self.origin_query = query;
     }
 
+    fn set_pre_execution_queries(&mut self, _pre_execution_queries: Option<&[String]>) {}
```
wangxiaoying, line 110: mark with [unimplemented](https://doc.rust-lang.org/std/macro.unimplemented.html) for unsupported sources?

connectorx/src/sources/postgres/mod.rs
```diff
@@ -199,7 +205,13 @@ where
     fn partition(self) -> Vec<Self::Partition> {
         let mut ret = vec![];
         for query in self.queries {
-            let conn = self.pool.get()?;
+            let mut conn = self.pool.get()?;
+
+            if let Some(queries) = &self.pre_execution_queries {
+                for query in queries {
```
wangxiaoying, line 211: rename the inner `query` (e.g., to `pre_query`) so that it won't be confused with the outer `query` variable?

connectorx/src/dispatcher.rs
```diff
@@ -30,7 +31,13 @@ where
     TP: Transport<TSS = S::TypeSystem, TSD = D::TypeSystem, S = S, D = D>,
 {
     /// Create a new dispatcher by providing a source, a destination and the queries.
-    pub fn new<Q>(src: S, dst: &'w mut D, queries: &[Q], origin_query: Option<String>) -> Self
+    pub fn new<Q>(
+        src: S,
+        dst: &'w mut D,
+        queries: &[Q],
+        origin_query: Option<String>,
+        pre_execution_queries: Option<&[String]>,
```
wangxiaoying, line 39: Instead of adding a new parameter to the constructor, maybe it is better to add a `set_pre_execution_queries` function? When using the dispatcher, we can:
1. let dispatcher = Dispatcher::new(...); -- remain the same
2. dispatcher.set_pre_execution_queries(...); --- optional, only set when needed
3. dispatcher.run() -- remain the same

I think it could be a way to avoid the breaking change. What do you think?

connectorx-python/connectorx/tests/test_postgres.py
```diff
@@ -1238,4 +1238,33 @@ def test_postgres_partition_with_orderby_limit_desc(postgres_url: str) -> None:
         },
     )
     df.sort_values(by="test_int", inplace=True, ignore_index=True)
```
wangxiaoying, line 1240: We probably also need to add a test for applying pre_execution_queries with partitioning enabled. To make the test case easier, we can use `query = [list of partitioned query]` to directly specify the query for each partition. For example:

```Python

query = [
        "SELECT CAST(name AS TEXT) AS name, CAST(setting AS INTEGER) FROM pg_settings WHERE name = 'statement_timeout'", 
        "SELECT CAST(name AS TEXT) AS name, CAST(setting AS INTEGER) FROM pg_settings WHERE name = 'idle_in_transaction_session_timeout'"
]
pre_execution_query = pre_execution_query = [
        "SET SESSION statement_timeout = 2151",
        "SET SESSION idle_in_transaction_session_timeout = 2252",
    ]

df = read_sql(postgres_url, query, pre_execution_query=pre_execution_query)

expected = pd.DataFrame(
        index=range(2),
        data={
            "name": pd.Series(["statement_timeout", "idle_in_transaction_session_timeout"], dtype="str"),
            "setting": pd.Series([2151, 2252], dtype="Int64"),
        },
)
```

wangxiaoying (commented):
Thank you @jsjasonseba  for the PR! I think it looks good in general

I left a few comments in the review. Please let me know if you have any questions!

connectorx/src/dispatcher.rs
jsjasonseba: Yeah, I think this a better approach. Thanks for the suggestion.

connectorx/src/sources/bigquery/mod.rs
jsjasonseba: Wow that's neat, just learned about this. Will implement them soon.

jsjasonseba:
Hi @wangxiaoying, I have finished implementing your comments. Feel free to check it at your convenience.

wangxiaoying:
Thank you @jsjasonseba ! I have merged the PR. I also  invited you as the collaborator of the project. Hope it will be easier for you to contribute in the future!

jsjasonseba:
Hi @wangxiaoying, you're welcome. Will do more contributions when I got the chance. 

Just wondering, when is the next release for this change? I am planning to support this use case in polars once it is released.

wangxiaoying:
> Just wondering, when is the next release for this change? I am planning to support this use case in polars once it is released.

I'm currently planning to do a stable release in mid February. But let me know if you want me to have an alpha release for this feature first so you can use it for polars immediately.

jsjasonseba:
I think stable release in mid February is fine. Thanks

wangxiaoying:
Hi @jsjasonseba , v0.4.2 has just released.

PR #731, opened 2025-01-21
Fix/mssql pre login token

Niivii:
Not bumping to latest versions because it requires some devs with breaking changes for UUID 1.0.
Might do them when I'm confident enough in Rust 🥲


wangxiaoying:
Thank you @Niivii !

Issue #736, opened 2025-01-26
Bug when Null are included inside an array (all types)


EricFecteau:
# Issue

When `null` are present in an array, the [try_get()](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/postgres/mod.rs#L449) gives a [WasNull](https://docs.rs/postgres/latest/postgres/types/struct.WasNull.html) PostgreSQL error. This is true for all the postgres types (Binary, CSV...). 

# Fix

Implementing `Vec<Option<T>>` instead of `Vec<T>` [here](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/postgres_arrow.rs#L71) and for all the postgres protocols [here](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/postgres/mod.rs#L465) (and a few other places) fixes the issue. It makes sense to always make it nullable, since arrow assumes it's always nullable with Arrow's LargeList being Vec<Option<T>> anyway.

I will fix this in my next PR.

Issue #735, opened 2025-01-25
Bug in TimeStamp with Time Zones for Arrow


EricFecteau:
# Issue

The Connector-X code converts timezone data using `UTC` around [here](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/destinations/arrow/arrow_assoc.rs#L210). But it does not seem to work correctly:

![Image](https://github.com/user-attachments/assets/32216053-f173-4ea5-a67f-9bcd712f88bf)

This is because, according to [arrow docs](https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.Timestamp), you have to have the `chrono-tz` feature enable to do this and Connector-X [does not have this enabled anywhere](https://github.com/sfu-db/connector-x/blob/main/Cargo.toml#L11).

> When feature chrono-tz is not enabled, allowed timezone strings are fixed offsets of the form “+09:00”, “-09” or “+0930”. When feature chrono-tz is enabled, additional strings supported by [chrono_tz](https://docs.rs/chrono-tz/latest/chrono_tz/) are also allowed, which include [IANA database](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) timezones.

# Fix

I am currently writing tests for all the types (at least all the types in PostgreSQL at this time) and I can fix this by either using `+00:00` instead of UTC (I confirmed it works) or by enabling the `chrono-tz` in my next PR. 

# Verification of understanding

Could you please also confirm my understanding. 

The `Timestamp` (without time zone) is mapped to `None` time zone in Arrow (through the `NaiveDateTime` in Chrono), because it is not meaningfully useful to think of it as UTC time, as explained in the Arrow docs: 

> Therefore, timestamp values without a timezone cannot be meaningfully interpreted as physical points in time, but only as calendar / clock indications (“wall clock time”) in an unspecified timezone.
>
> For example, the timestamp value 0 with an empty timezone string corresponds to “January 1st 1970, 00h00” in an unknown timezone: there is not enough information to interpret it as a well-defined physical point in time.
>
> One consequence is that timestamp values without a timezone cannot be reliably compared or ordered, since they may have different points of reference. In particular, it is not possible to interpret an unset or empty timezone as the same as “UTC”.  

And while the PostgreSQL docs don't explicitly mention this, they also don't call their `Timestamp` as having `UTC` time, but they [definitely call](https://www.postgresql.org/docs/current/datatype-datetime.html) their `TimestampTz` time `UTC`

[...]

Issue #688, opened 2024-09-26
Int dtypes are unnecessarily big


Happily-Coding:
#### What language are you using?

Python

#### What version are you using?

0.3.3

#### What database are you using?

MySQL

#### What dataframe are you using?

Polars

#### Can you describe your bug?

Connectorx pointlessly transforms smaller db dtypes into int64, leading to higher memory usage.
See [mysql mappings]( https://github.com/sfu-db/connector-x/blob/a202e7285e58a0c8bc30a69e3afd653a8b4f519b/connectorx/src/transports/mysql_arrow.rs) or [MSSQL mappings]( https://github.com/sfu-db/connector-x/blob/a202e7285e58a0c8bc30a69e3afd653a8b4f519b/connectorx/src/transports/mssql_arrow.rs#L27)

#### What are the steps to reproduce the behavior?
Perform a query to mysql using connectorx to a table containing a smaller dtype in a column.
The df will have int64 dtypes in all numeric dtypes, as specified by the incorrect transports linked above.

For example with polars which respectes the dtypes provided by connectorx:
```
import polars as pl
mysql_uri = f'{mysql_db_type}://{mysql_username}:{mysql_password}@{mysql_host}:{mysql_port}/{mysql_database_name}'
test_dataset = pl.read_database_uri(query, self.mysql_uri)
```

#### What is the error?
There will be no error message, just incorret dtypes.

PR #743, opened 2025-01-30
Update int/float size for PostgreSQL, add PostgreSQL tests and fix various bugs

EricFecteau:
# Goal

The goal of this pull request is to:
* Make the size of the `int` and `float` smaller (e.g. don't convert all `i16` to `i64`) --  this adds the framework for #688 for Postgres -> Polars, but not mysql -> Polars)
* Add tests for these types (and all other types) - mostly in Postgres for now

# Comments

* While writing more tests, I changed the PostgreSQL examples, so had to update the relevant tests in Rust and Python.
* I wrote tests for the the 4 protocols for PostgreSQL: Binary, CSV, Cursor and Simple
* Made clippy happy in various places (mostly PostgreSQL)

# Bugs

While writing tests, I encountered a few bugs and solved them:
* #736 `Null` within an array would fail in PostgreSQL! I implemented this to work in Rust. I did not implement this for Pandas (but kept it as-is, with an `unimplemented!`
* #735 The `UTC` time being used in the crate was not recognized by Arrow, due to the lack of the `chrono-tz` feature. Fixed by moving to `+00:00` instead.
* The PostgreSQL CSV and Simple protocol did not implement millisecond for time, so I implemented it.
* Bool was incorrectly implemented for the PostgreSQL Simple protocol, so I fix it!


connectorx/src/transports/postgres_arrow.rs
```diff
@@ -44,36 +44,36 @@ macro_rules! impl_postgres_transport {
...
+                { Timestamp[NaiveDateTime]           => Date64Micro[NaiveDateTimeWrapperMicro] | conversion option }
+                { Date[NaiveDate]                    => Date32[NaiveDate]                      | conversion auto   }
+                { Time[NaiveTime]                    => Time64Micro[NaiveTimeWrapperMicro]     | conversion option }
+                { TimestampTz[DateTime<Utc>]         => DateTimeTzMicro[DateTimeWrapperMicro]  | conversion option }
+                { UUID[Uuid]                         => LargeUtf8[String]                      | conversion option }
+                { Char[&'r str]                      => LargeUtf8[String]                      | conversion none   }
+                { ByteA[Vec<u8>]                     => LargeBinary[Vec<u8>]                   | conversion auto   }
+                { JSON[Value]                        => LargeUtf8[String]                      | conversion option }
+                { JSONB[Value]                       => LargeUtf8[String]                      | conversion none   }
+                { BoolArray[Vec<Option<bool>>]       => BoolArray[Vec<Option<bool>>]           | conversion option }
```
wangxiaoying, line 68: Here and the following transport `Vec<Option<T>> => Vec<Option<T>>` can be set to [`conversion auto`](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/macros.rs#L286) for the first appearance of `T` since the two types are exactly the same; and set to [`conversion none`](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/macros.rs#L320) for the later ones to avoid redundant implementation. 

After doing this, we can just remove the following manual implementation of `TypeConversion` for these types.

connectorx-python/src/pandas/transports/postgres.rs
```diff
@@ -79,12 +79,98 @@ impl<'py, P, C> TypeConversion<HashMap<String, Option<String>>, String>
     }
 }
 
-impl<'py, P, C> TypeConversion<Vec<Decimal>, Vec<f64>> for PostgresPandasTransport<'py, P, C> {
-    fn convert(val: Vec<Decimal>) -> Vec<f64> {
+impl<'py, P, C> TypeConversion<Vec<Option<Decimal>>, Vec<f64>>
```
wangxiaoying, line 82: We probably should also make the array type in pandas nullable (in either this PR or another PR), as it is supported in pandas.

```Python
import pandas as pd

data = [[1, [1,2,None,4]], [2, [1,2]], [3, [0,2,4]]]
df = pd.DataFrame(data, columns=['id', 'scores'])

df
"""
   id           scores
0   1  [1, 2, None, 4]
1   2           [1, 2]
2   3        [0, 2, 4]
"""
```

wangxiaoying (commented):
Hi @EricFecteau , thanks for the PR! It looks good. I only left a few comments.

connectorx/src/transports/postgres_arrow.rs
EricFecteau: Thank you! It has been fixed!

connectorx-python/src/pandas/transports/postgres.rs
EricFecteau: I completely agree that it should be done, but in another PR! I partially implemented it, but my background is in SAS, R and Rust, so I have low familiarity with Pandas, and got stuck. Likely just need to do more research, but  decided that the it would be best to do that in it's own PR.

wangxiaoying: Yeah, it makes sense to do in a separate PR.

Issue #773, opened 2025-03-17
Support for building in Docker on Mac M1?


ssharkey-rsg:
#### What language are you using?

Python

#### What version are you using?

0.4.2

#### What database are you using?

PostgreSQL

#### What dataframe are you using?

Pandas, Arrow

#### Can you describe your bug?

Cannot build in a linux (ubuntu) container on a Mac M1 (arm)

#### What are the steps to reproduce the behavior?

any app with connector-x == 0.4.2 in the pyproject.toml

##### Database setup if the error only happens on specific data or data type

N/A

##### Example query / code

docker compose build

#### What is the error?

hint: You're on Linux (`manylinux_2_36_aarch64`), but `connectorx` (v0.4.2) only has wheels for the following platforms: `manylinux_2_28_x86_64`, `macosx_10_7_x86_64`, `macosx_11_0_arm64`, `win_amd64`

This looks like 2 problems - 1) I'm using glibc 2.36, not 2.28, and 2) there is no aarch64 wheel.

pangjunrong:
Related to #240.

There has been some attempts to build `connector-x` for aarch64 but none has been successful thus far. If you build from source without `integrated-auth-gssapi` as a dependency, it will work for glibc 2.28 but I haven't tried it with 2.36.

wangxiaoying:
Resolved by #781 (support from version 0.4.3a1).

Issue #673, opened 2024-08-21
Enable Linux arm64 release wheels


samster25:
Our users in Daft rely on connector-x to read SQL from their databases. Our user base heavily uses graviton machines on AWS which are linux + arm64. 

We currently maintain a separate dependency path for linux/arm64 since the last published one was `0.2.3`.

Daft is also built in rust + pyo3 and we [publish linux arm64 wheels](https://github.com/Eventual-Inc/Daft/blob/b3f5260740cb6ab5013671b31d6a8a1e89fc4002/.github/workflows/python-publish.yml#L34). Happy to contribute!

https://github.com/Eventual-Inc/Daft/blob/b3f5260740cb6ab5013671b31d6a8a1e89fc4002/requirements-dev.txt#L61

vc1492a:
+1

wangxiaoying:
Resolved by #781 (support from version 0.4.3a1).

Issue #535, opened 2023-09-02
Unable to find connectorx version on linux/aarch64


MatsMoll:
#### What language are you using?

**Python**

#### What version are you using?

Trying to install 0.3.2

#### Can you describe your bug?

When installing connectorx on a `linux/aarch64` (docker on an M2 Mac) architecture, `pip` not be able to find any versions to install.

#### What are the steps to reproduce the behavior?

I have a Docker image built similar to the following.

```Dockerfile
FROM python:3.10.5-slim
RUN pip install connectorx==0.3.2
```

If this is build on an M2 Mac will it fail.

#### What is the error?

The error is the following:

```
67.29   RuntimeError
67.29
67.29   Unable to find installation candidates for connectorx (0.3.2)
67.29
67.29   at /usr/local/lib/python3.10/site-packages/poetry/installation/chooser.py:73 in choose_for
67.29        69│
67.29        70│             links.append(link)
67.29        71│
67.29        72│         if not links:
67.29     →  73│             raise RuntimeError(f"Unable to find installation candidates for {package}")
67.29        74│
67.29        75│         # Get the best link
67.29        76│         chosen = max(links, key=lambda link: self._sort_key(package, link))
67.29        77│
```

#### Why it fails
When running `pip index versions connectorx` in an interactive docker image on a `linux/aarch64` platform, will the following be returned.

```bash
WARNING: pip index is currently an experimental command. It may be removed/changed in a future release without prior warning.
connectorx (0.2.3)
Available versions: 0.2.3
```

The only difference between `0.2.3` and later versions is that it contains a source distribution.
Ideally something like a dedicated `linux/aarch64` image could have been built, but providing a source distribution would also be nice.

#### Potential fix?
Looking at the CI for the `0.2.3` version do I see that an extra step is added.

```
- name: Persist artifacts source code
        uses: actions/upload-artifact@v2
        if: matrix.python-version == 39
        with:
          name: "linux-${{ matrix.python-version }}"
          path: connectorx-python/dist/*.tar.gz
```
However, this have been removed in later versions, which makes me assume that this was not working totally as expected?

Jandelson:
I have the same issue.

vinibacc:
me too, i have same issue.

lBilali:
I'm having a similar issue when trying to install it on Windows 11 with Python 12.2 on an venv 
running `py -m pip install connectorx==0.3.2`

I get 
```
[...]

Issue #386, opened 2022-10-30
Add linux aarch64 support


leodiegues:
Hi!

I've been really struggling to install the package in an EC2 instance that happens to have an aarch64 processor and is running ubuntu. Are there any plans to add aarch64 support for linux machines in the future?

Regards

wangxiaoying:
Hi @leonardodiegues , currently we use github action and github hosted runners to manage our release process. So we cannot support aarch64 linux since it is not in the list [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners). For linux aarch64, you may need to build your own wheel file like mentioned in [this doc](https://sfu-db.github.io/connector-x/install.html#build-from-source-code)

leonardodiegues:
Thank you for your answer. I'll follow the docs build instructions!

Sirsirious:
The build instructions do not work out of the box, especially as it points to ever changing Rust nightly versions.

Lacking official Linux aarch64 support hinders the adoption of this library, as it is basically the version that every developer using a mac M1 and docker will end up needing for local testing. Could we have better instructions/support for that?

wangxiaoying:
Hi @Sirsirious , thanks for the comment. For mac M1 we do have precompiled wheel file support. You can just run `pip install connectorx` to install it. So only linux users with aarch64 processor need to build connectorx from the source. We also tried to cross compile it in our release action (like what we did for mac m1) but still has some issues. A good news is that we have switched to rust stable since the GAT is stabilized !

duvenagep:
Hi @wangxiaoying, 

I am a big fan of this library, use it a lot for EDA in Jupyter Notebooks with Polars with big datasets >= 200GB. I want to now use it in a production env on Airflow! Unfortunately my whole team uses mac M1 and every team member builds a local a local dev airflow environment from a docker-compose file on a shared github repo! 

I currently use polars but read in the data using pd.read_sql_query and then transform the data to a polars dataframe with  pl.from_pandas! I would prefer to use connectorX with Polars! 

Below is the error log trying to install ConnectorX. Currently building from source for the whole team is currently not an option.

Is there any update on the support for aarch64 support?


```
 Downloading polars-0.16.5-cp37-abi3-manylinux_2_17_aarch64.manylinux2014_aarch64.whl (13.3 MB)
[...]

PR #781, opened 2025-03-18
feat: Linux ARM Release CICD for `glibc 2.35`

pangjunrong:
This change re-introduces the Linux ARM Build process to the release workflow and seeks to address the long-standing issue of missing `connector-x` wheels for Linux ARM in #240, #386, #506, #535, #673, #724 and more recently, #773.

With the introduction of GitHub ARM-based runners earlier this year in public preview, it is now possible to build our distribution natively in an ARM-based Ubuntu 22.04 environment, which should allow compatibility with OSes using `glibc 2.35` and above.

A test run can be found in my fork here:
https://github.com/pangjunrong/connector-x/actions/runs/13912310291

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,67 +7,67 @@ on:
       - release
 
 jobs:
-  # linux:
-  #   runs-on: ubuntu-latest
-  #   container: quay.io/pypa/manylinux_2_28_x86_64
-  #   strategy:
-  #     matrix:
-  #       python-version: [[310, "3.10"], [311, "3.11"], [312, "3.12"], [313, "3.13"]]
-  #   steps:
-  #     - uses: actions/checkout@v4
+  linux:
+    runs-on: ubuntu-latest
+    container: quay.io/pypa/manylinux_2_28_x86_64
+    strategy:
+      matrix:
+        python-version: [[310, "3.10"], [311, "3.11"], [312, "3.12"], [313, "3.13"]]
+    steps:
+      - uses: actions/checkout@v4
 
-  #     - name: Set python version
-  #       run: |
-  #         echo "/opt/python/cp${{ matrix.python-version[0] }}-cp${{ matrix.python-version[0] }}/bin" >> $GITHUB_PATH
+      - name: Set python version
+        run: |
+          echo "/opt/python/cp${{ matrix.python-version[0] }}-cp${{ matrix.python-version[0] }}/bin" >> $GITHUB_PATH
 
-  #     - uses: actions-rs/toolchain@v1
-  #       with:
-  #         toolchain: stable
-  #         components: rustfmt
-  #         target: aarch64-unknown-linux-gnu
-  #         default: true
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: rustfmt
+          target: aarch64-unknown-linux-gnu
+          default: true
 
-  #     - uses: extractions/setup-just@v2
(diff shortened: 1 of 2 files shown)


pangjunrong:
Hey @wangxiaoying whenever you're free, could you take a look at this PR to re-enable Linux ARM releases for newer distributions?

vc1492a:
Sorely needed by our team, this would be a great addition! Thanks @pangjunrong!

.github/workflows/release.yml
```diff
@@ -69,53 +69,78 @@ jobs:
...
+
+      - name: Build Wheel (Native)
+        run: |
+          maturin build -m connectorx-python/Cargo.toml --target aarch64-unknown-linux-gnu -i python${{ matrix.python-version }} --release
+        env:
+          SQLITE3_STATIC: 1
+          KRB5_INCLUDE_DIR: /usr/include
+          KRB5_LIB_DIR: /usr/lib/aarch64-linux-gnu
+
+      - name: Copy J4RS Dependencies
```
wangxiaoying, line 129: j4rs should no longer be a dependency by default. We removed it from the release script [previously](https://github.com/sfu-db/connector-x/commit/8090d2fd15a635af1f3ebbbe8539b41861c4337d) for all the platforms.

We can remove `Copy J4RS Dependencies` and `Rebuild Wheel (Native)` here as well.

wangxiaoying:
Hi @pangjunrong , thank you so much for the PR! 

I added a minor comment related to the `j4rs` dependency. After fixing it we can merge it and have an alpha release for this!

.github/workflows/release.yml
pangjunrong: Thanks @wangxiaoying, I've removed both steps and did a rerun on my fork for a sanity check and it looks good to go!

The new test run with J4RS removed can be found in my fork here: https://github.com/pangjunrong/connector-x/actions/runs/14243748048

vc1492a: This is **huge** @wangxiaoying! 🚀

wangxiaoying:
I released a new alpha version, which can be installed by `pip install connectorx==0.4.3a1`. I don't have a linux arm instance. Can anyone try whether it works?

pangjunrong:
> I released a new alpha version, which can be installed by `pip install connectorx==0.4.3a1`. I don't have a linux arm instance. Can anyone try whether it works?

Hey @wangxiaoying, thanks for the quick turnaround! I've tested it in a Docker ARM64 Ubuntu 22.04 setup together with PostgresDB and it works (i.e., I am able to retrieve test data from the DB into a dataframe). The run and its logs can be found here:
https://github.com/pangjunrong/connector-x/actions/runs/14260363710

Separately and as expected, the library is not found for Ubuntu 20.04 and cannot be installed as it runs an older version of `glibc` (i.e., 2.31 instead of the minimum 2.35). Perhaps it will be prudent to update the documentation on this nuance, to provide more clarity on which Linux ARM64 distributions can run `connector-x` now. I do the documentation update accordingly and submit a PR accordingly :smile:

vc1492a:
FWIW, I built the my company's platform locally on macOS with this pre-release and it works, too! ❗ 🚀

wangxiaoying:
Thank you @pangjunrong @vc1492a for testing this!

@pangjunrong Yes, we can definitely add a clarification in the documentation about the glibc version. I have invited you as a collaborator of this repo. Hopefully it will make it easier for you to contribute!

I will release a stable version next week!

Issue #783, opened 2025-03-31
[MSSQL] Database name is not decoded from URL encoded connection string


tvandelooij:
#### What language are you using?

Rust

#### What version are you using?

`v0.4.0`

#### What database are you using?

Microsoft SQL Server

#### What dataframe are you using?

Arrow

#### Can you describe your bug?

In MS SQL Server, database names can contain a space, e.g. `My Database` or `[My Database]`. After a `url::URL` struct is created from the `connector-x` connection string, the database name (`path`) is not decoded which results in a database name that cannot be found.

`My%20Database` instead of `"My Database"`.

#### What are the steps to reproduce the behavior?

Minimal example using only the `url` and `urlencoding` crates:

```rustfn main() {
    let url_str =
        "mssql://sa:Welcome_123@localhost:1433/My Database?xcprotocol=binary&encrypt=false";
    let url = url::Url::parse(url_str).unwrap();

    println!("{}", &url.path()[1..]);
    println!("{:?}", urlencoding::decode(&url.path()[1..]).unwrap());
}
```

#### What is the error?

```
[2025-03-31][09:43:07][tiberius::client::connection][WARN] Trusting the server certificate without validation.
[2025-03-31][09:43:07][tiberius::client::connection][INFO] TLS handshake successful
[2025-03-31][09:43:07][tiberius::tds::stream::token][ERROR] Cannot open database "My%20Database" requested by the login. The login failed. code=4060
```

PR #784, opened 2025-03-31
URL decode database name for MSSQL

tvandelooij:
Resolves https://github.com/sfu-db/connector-x/issues/783


wangxiaoying:
Thank you @tvandelooij for the PR!

Issue #786, opened 2025-04-04
pd.core.arrays.DatetimeArray Issue


sssaha:
#### What language are you using?

**Python**

#### What version are you using?

0.4.2.

#### What database are you using?

SQLITE

#### What dataframe are you using?

Pandas

#### Can you describe your bug?

Not compatible with Latest pandas 2.2.3 - in the SQL database has datetime

#### What is the error?

    df = cx.read_sql(conn, query)
  File "C:\ADO\mdct\venvp313\Lib\site-packages\connectorx\__init__.py", line 394, in read_sql
    df = reconstruct_pandas(result)
  File "C:\ADO\mdct\venvp313\Lib\site-packages\connectorx\__init__.py", line 478, in reconstruct_pandas
    pd.core.arrays.DatetimeArray(block_data), placement=binfo.cids
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "C:\ADO\mdct\venvp313\Lib\site-packages\pandas\core\arrays\datetimelike.py", line 1950, in __init__
    warnings.warn(
    ~~~~~~~~~~~~~^
        # GH#55623
        ^^^^^^^^^^
    ...<3 lines>...
        stacklevel=find_stack_level(),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
FutureWarning: DatetimeArray.__init__ is deprecated and will be removed in a future version. Use pd.array instead.

wangxiaoying:
Hi @pangjunrong , thanks for the investigation!

The performance of the array construction using `_from_sequence` looks pretty impressive! I think we could switch to it for all the ExtensionArrays.

As for `make_block`, we probably can wait and see how `pyarrow` deal with this.

sssaha:
Any speed gain is always preferable - that is the reason we are using connectorx than panda read SQL. I can wait to upgrade my python version (It seems like Connectorx 0.3.3 is not compatible with Python 3.13)

PR #795, opened 2025-04-13
refactor: Optimize DataFrame Reconstruction & Update Docs for Linux ARM64 Release

pangjunrong:
This change optimizes the reconstruction of Pandas DataFrames using the `_from_sequence` methods for ExtensionArrays, which run on bulk conversion routines using vectorized NumPy operations while avoiding the extra layers of validation in the general constructor.

It also uses the `_from_mgr` method for DataFrame construction using BlockManager, which assumes the BlockManager is already in a valid state and skips the validation and setup overhead that the regular pd.DataFrame(BlockManager) constructor performs.

This also resolves the warnings mentioned in #786.

Based on a simple generic benchmark between the typical class initialization and use of `_from_sequence`, the performance gain seems to be consistent & profound across BooleanArray, DatetimeArray & IntegerArray. However, I will need help for additional verification on this as I don't fully understand the full picture of how Pandas Internals operate at the lower level.

BooleanArray
DatetimeArray
IntegerArray

Separately, the installation guides are updated to inform users of `connectorx==0.4.3`'s general availability for Linux ARM64 distributions running on glibc 2.35 & later and the use of `connectorx=0.2.3` for older distributions not covered by our new build process.

### DIFF
--- a/connectorx-python/connectorx/__init__.py
+++ b/connectorx-python/connectorx/__init__.py
@@ -459,20 +459,16 @@ def reconstruct_pandas(df_infos: _DataframeInfos) -> pd.DataFrame:
                 pd.core.internals.make_block(block_data, placement=binfo.cids)
             )
         elif binfo.dt == 1:  # IntegerArray
-            integer_array = pd.core.arrays.IntegerArray._from_sequence(block_data[0])
-            integer_array._mask = block_data[1]
             blocks.append(
                 pd.core.internals.make_block(
-                    integer_array,
+                    pd.core.arrays.IntegerArray(block_data[0], block_data[1]),
                     placement=binfo.cids[0],
                 )
             )
         elif binfo.dt == 2:  # BooleanArray
-            bool_array = pd.core.arrays.BooleanArray._from_sequence(block_data[0])
-            bool_array._mask = block_data[1]
             blocks.append(
                 pd.core.internals.make_block(
-                    bool_array,
+                    pd.core.arrays.BooleanArray(block_data[0], block_data[1]),
                     placement=binfo.cids[0],
                 )
             )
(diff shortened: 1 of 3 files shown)


connectorx-python/connectorx/__init__.py
```diff
@@ -459,23 +459,27 @@ def reconstruct_pandas(df_infos: _DataframeInfos) -> pd.DataFrame:
...
+            integer_array._mask = block_data[1]
             blocks.append(
                 pd.core.internals.make_block(
-                    pd.core.arrays.IntegerArray(block_data[0], block_data[1]),
+                    integer_array,
                     placement=binfo.cids[0],
                 )
             )
         elif binfo.dt == 2:  # BooleanArray
+            bool_array = pd.core.arrays.BooleanArray._from_sequence(block_data[0])
```
wangxiaoying, line 471: From what I understand from the pandas source code ([_from_sequence](https://github.com/pandas-dev/pandas/blob/main/pandas/core/arrays/masked.py#L145), [coerce_to_array](https://github.com/pandas-dev/pandas/blob/main/pandas/core/arrays/boolean.py#L160)), it seems we will have an extra mask array constructed by this `_from_sequence` step, which will then be discarded and replaced by our mask array like in this example:
image

And also it seems to directly [call the constructor](https://github.com/pandas-dev/pandas/blob/main/pandas/core/arrays/masked.py#L146) of the `BooleanArray` anyway. I'm wondering why this `_from_sequence` approach is still faster than the old `BooleanArray(data, mask)` approach as it seems to only include the overhead of mask construction. I'm I missing something here?

wangxiaoying:
Hi @pangjunrong , thank you so much for the PR!

The `DatetimeArray` part looks great to me! I have a question about the IntegerArray and BooleanArray when I check on the pandas code as I left in the review.

connectorx-python/connectorx/__init__.py
pangjunrong: Thanks for verifying this, you're right that for BooleanArray & IntegerArray, the additional overhead of mask construction would mean that the `from_sequence` approach would be slower — I seem to have missed that out when benchmarking the different array types.

I will reflect this accordingly!

PR #801, opened 2025-05-05
Bump Arrow to 0.54

EricFecteau:
Arrow 0.53 (current used by connectorX) does not play well with any crates needing Arrow 0.55 (the latest version of Arrow). For example, datafusion has updated to Arrow 0.55 and is no longer working with Arrow 0.53. For example, a crate that imports the latest version of `connectorx` and of `datafusion` can not compile. See issue here: https://github.com/apache/arrow-rs/issues/7410

```
[dependencies]
connectorx = { version = "0.4", features = ["src_postgres", "dst_arrow"] }
datafusion = "47"
```

Arrow 0.54 seems to play well with 0.55 and 0.53, so that is why I bumped it up to this version only. I fear bumping it up to 0.55 will cause issues with other crate features I can not test (e.g. older version of datafusion).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,53 +190,53 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "arrow"
-version = "53.4.0"
+version = "53.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eaf3437355979f1e93ba84ba108c38be5767713051f3c8ffbf07c094e2e61f9f"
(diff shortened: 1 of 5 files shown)


EricFecteau:
Had to bump `datafusion` also because same error as https://github.com/sfu-db/connector-x/pull/771 (but opposite direction). Bumped it 0.46, the one that has the `arrow 54` version. All arrows in this crate now match and should be maximally compatible with other crates for now.

EricFecteau:
Also confirmed that this branch solves the issue of compatibility with other crates:

```
[dependencies]
connectorx = { git = "https://github.com/sfu-db/connector-x.git", branch = "update_arrow", features = ["src_postgres", "dst_arrow"] }
datafusion = "47"
```

wangxiaoying:
Thanks for the PR @EricFecteau , it looks good to me.

Issue #798, opened 2025-04-15
Add Postgres Inet types


abstractqqq:
#### Describe your feature request

Inet types are essentially data types for IP addresses. I know that arrow doesn't have this type yet, but can we at least add an option to cast inet to strings? 

[You can find more about postgres inet types here.](https://www.postgresql.org/docs/current/datatype-net-types.html)

Please describe the behavior you want and the motivation.

For people doing analytics, it is common to run a query like
```python
select * from table limit 100
```
And if there is a inet type in the schema, it panics now with the error "not implemented inet". I think for inet types it is very reasonable to cast them to string.

PR #808, opened 2025-05-23
feat(postgres): add support for Inet type in Postgres integration

holicc:
[#798](https://github.com/sfu-db/connector-x/issues/798)


wangxiaoying:
The code looks good to me. Thanks @holicc again!

Maybe later we can add this type to other destinations as well, including [pandas](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/pandas/transports/postgres.rs) and [arrow_stream](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/postgres_arrowstream.rs).

PR #828, opened 2025-07-19
Fix `future-incompat` report

EricFecteau:
When building, `connector-x` provides a `future-incompat` warning (see here: https://github.com/sfu-db/connector-x/actions/runs/16383175035/job/46298417726?pr=827#step:12:534) for the crate `connection-string v0.1.13`. This was a dependency of  the version of `tiberius` being used by `connector-x`. I updated `tiberius` to the latest version, and updated the related crates `bb8` and `bb8-tiberius`. This also solved the incompatibility of `uuid` version 1 (removing the `uuid_old` crate. 

**Warning:** I don't have an MS SQL server to verify that this did not cause any issues. Uncertain what to do about that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


EricFecteau:
Ah yeah, that makes sense! Will close for now.

PR #844, opened 2025-08-26
fix: feature gate utils fn using optional `rust_decimal` dep

lmmx:
Hi there, hope this small PR is not overlooking anything: I tried to build the trunk and it failed due to one dependency miss, the `rust_decimal` import is actually an optional dependency.

The utils function it is used in is only called by code gated by the `arrow` feature flag, so it is equivalent to feature-gate the util too. I moved the `use rust_decimal::Decimal;` line down into a direct reference to it in the parameter, to avoid repeating the feature gate directive, and likewise for `anyhow::Result` (to avoid the unused import warning).

- It entered the code base 3 months ago here: https://github.com/sfu-db/connector-x/commit/c7d6ee3b6017f2c14ae243859e770b7270ebdc9b


lmmx:
- This is also in #845, feel free to accept that one instead if it's ready or merge separately

PR #853, opened 2025-10-10
Add python 3.14 spport

Vincenthays:


surister:
@Vincenthays Do you have a link to the failing CI?

Jazzinghen:
I can try locally as well, however it seems like it might be necessary to explicitly install `libffi-dev` on the runner?
It sounds strange to me because it should have been an issue even in the past.
I'll see if I can do something as well.

Vincenthays:
it worked 🎉 
https://github.com/Vincenthays/connector-x/actions/runs/19861980055/job/56914241598#logs
I had to upgrade numpy, pytest and pytest-benchmark

Jazzinghen:
Woah!
I have been also trying to make it work, however I might have gone a bit in a strange way.
One thing I am wondering is why it was OK to not have `python-dev` before. It's as if something else installed it :thinking:

PR #856, opened 2025-10-17
build aarch64 with manylinux2_28

davidhewitt:
Uses the `manylinux_2_28` container to run the aarch64 build (on an arm runner natively).

cc @pangjunrong you noted that this had problems building with the `manylinux_2_28` container. From the test run on my fork, this worked fine: https://github.com/davidhewitt/connector-x/actions/runs/18595020541

Maybe I missed something?

This allows the aarch64 build to match the x86_64 one of glibc 2.28.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -201,7 +201,6 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-          architecture: x64
 
       - uses: actions/download-artifact@v5
         with:


.github/workflows/release.yml
```diff
@@ -66,70 +72,7 @@ jobs:
...
-        env:
-          SQLITE3_STATIC: 1
-          KRB5_INCLUDE_DIR: /usr/include
-          KRB5_LIB_DIR: /usr/lib/aarch64-linux-gnu
-
-      - name: Upload Artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: "linux-arm-${{ matrix.python-version }}"
+          name: "linux-${{matrix.platform.arch}}-${{ matrix.python-version[1] }}"
```
wangxiaoying: Since the name is changed from `ubuntu-latest-xxx` to `linux-x86_64-xxx`. The named referred in the `verify` task should also be updated. Otherwise it cannot find the artifact: https://github.com/davidhewitt/connector-x/actions/runs/18595020541/job/53024260726#step:4:16

PR #857, opened 2025-10-18
feat(oracle): support schema query parameter in connection URL

kongscn:
Add support for specifying Oracle schema via URL query parameter like`?schema=abc`. When 'schema' is provided, it's automatically set as the current schema for all connections in the pool.


wangxiaoying:
Thanks @kongscn for the PR! The code looks good to me. Can you also add this to the [documentation](https://github.com/sfu-db/connector-x/blob/main/docs/databases/oracle.md)? I can merge it after that.

PR #871, opened 2025-12-17
feat: enhance oracle python tests using Testcontainers library

dbascoules:
I suggest using [Testcontainers](https://testcontainers.com) for ease of testing. Here is an example for Oracle testing in Python.
Doing so we could perhaps reproduce some bugs like #866 or https://github.com/dlt-hub/dlt/issues/3491 more easily and fix them.
@wangxiaoying what do you think ?


wangxiaoying:
Thanks @dbascoules for the suggestion. I think it is a great idea! It will make the test for everyone much easier, as currently our CI tests cannot be really executed without deploying the corresponding databases/container locally. We can use Testcontainers for Oracle first and extend to other databases later (can replace the existing docker deployment in the CI workflow for postgres and mysql [here](https://github.com/sfu-db/connector-x/blob/main/.github/workflows/ci.yml#L18) ).

Other than the container itself, we probably also need to install corresponding dependencies in the CI workflow to [enable the automatic CI test](https://github.com/sfu-db/connector-x/actions/runs/20349365049/job/58469655844?pr=871).

PR #885, opened 2026-01-27
fix: correctly detect MySQL BINARY/VARBINARY types using BINARY_FLAG

yuansuentw:
## Summary

MySQL BINARY/VARBINARY types were misidentified as CHAR/VARCHAR because the type system didn't check the BINARY_FLAG to distinguish between text and binary types sharing the same MySQL type code.

This caused ConnectorX to attempt converting binary data to UTF-8 strings, resulting in panics like `Could not retrieve Option<String>: Couldn't convert the value Bytes(...) to a desired type`.

## Technical Details

MySQL wire protocol uses shared type codes (MYSQL_TYPE_STRING, MYSQL_TYPE_VAR_STRING) for both text and binary types. The BINARY_FLAG in column metadata is required to distinguish them, which was previously not checked.

This fix adds BINARY_FLAG detection in `typesystem.rs` to correctly map:
- BINARY/VARBINARY → Vec<u8> (binary data)
- CHAR/VARCHAR → String (text data)

## Testing

Added comprehensive test coverage:
- Python integration tests with BINARY/VARBINARY columns
- Rust unit tests for BINARY_FLAG detection logic (no database required)
- Updated test schema in `scripts/mysql.sql`

### DIFF
--- a/connectorx-python/src/pandas/transports/oracle.rs
+++ b/connectorx-python/src/pandas/transports/oracle.rs
@@ -2,12 +2,12 @@ use crate::errors::ConnectorXPythonError;
 use crate::pandas::destination::PandasDestination;
 use crate::pandas::typesystem::PandasTypeSystem;
 use chrono::{DateTime, NaiveDateTime, Utc};
-use rust_decimal::prelude::*;
 use connectorx::{
     impl_transport,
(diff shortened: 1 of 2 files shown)


wangxiaoying:
Thanks @yuansuentw for the bug fix! It looks great to me!

PR #896, opened 2026-03-28
Feature/allow redshift iam or saml login

antivirak:
This lets user provide redshift connection string with parameters to login through SAML; currently supporting Ping.
https://docs.aws.amazon.com/redshift/latest/mgmt/options-for-providing-iam-credentials.html
The connection string needs to start with redshift-iam://, so it does not get converted to postgres:// before reaching source_router.
Tested from polars:
pip3 install pyarrow polars connectorx
```
>>> import polars as pl
>>> pl.read_database_uri('select col from schema.table limit 2', 'redshift-iam://user:password@host:port/db?clusterId=cluster-name&IdP_Port=443&IdP_Host=pingfed.example.com&Plugin_Name=com.amazon.redshift.plugin.PingCredentialsProvider&Preferred_Role=arn:aws:iam::123:role/role-name', protocol='cursor')
shape: (2, 1)
...
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,12 +93,12 @@ MODIN_ENGINE=dask
 * Define the conversion between the new source and existing destinations
   * Examples can be found [here](https://github.com/sfu-db/connector-x/tree/main/connectorx/src/transports) and [here](https://github.com/sfu-db/connector-x/tree/main/connectorx-python/src/pandas/transports)
 * Make the new source visable to destinations, including:
-  * Add the source to the [source_router](https://github.com/sfu-db/connector-x/blob/main/connectorx-python/src/source_router.rs)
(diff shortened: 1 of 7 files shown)


wangxiaoying:
Hi @antivirak , thanks for the PR and sorry for the late reply!

The code looks good to me. Can you also update the document [here](https://github.com/sfu-db/connector-x/blob/main/docs/databases/postgres.md) (or adding a new *.md file for redshift only)? We can merge this PR after this.

antivirak:
Hi, I added docs to the Postgres.md. I think it is not necessary to create separate redshift document, on the other hand I'm keeping it short to not hijack the Postgres one too much

wangxiaoying:
Thank you @antivirak ! The PR is merged.

PR #899, opened 2026-04-01
fix: prevent division-by-zero and integer overflow in partition()

hieuit095:
## Summary

Harden `partition()` in `connectorx/src/partition.rs` against arithmetic panics and invalid partition ranges:

- **Reject `num == 0`**: Return an explicit error instead of panicking on division by zero.
- **Reject invalid/inverted ranges**: Return an explicit error when `max < min`.
- **Checked range calculation**: Use checked arithmetic for `max - min + 1`.
- **Checked partition boundaries**: Use `checked_mul` and `checked_add` instead of saturating/clamping arithmetic.
- **Protect final upper bound**: Return an explicit error when `max + 1` cannot be represented.
- **Preserve partition count semantics**: When the requested partition count exceeds the number of integer values in the range, zero-sized partitions remain valid rather than introducing a new error.
- **Tests**: Cover zero partitions, inverted ranges, arithmetic overflow, upper-bound overflow, and the small-range/many-partitions compatibility case.

### DIFF
--- a/connectorx/src/partition.rs
+++ b/connectorx/src/partition.rs
@@ -84,13 +84,60 @@ pub fn partition(part: &PartitionQuery, source_conn: &SourceConn) -> OutResult<V
         )),
     };
 
-    let partition_size = (max - min + 1) / num;
+    // Guard against division by zero (num must be > 0)
+    if num == 0 {
+        throw!(anyhow!("partition count (num) must be greater than zero"));
+    }
+
+    // Guard against empty or inverted range (max < min).
+    // When max < min, the range is empty or invalid; return a single empty query.
+    if max < min {
+        // Return one partition spanning the invalid range so callers receive a result
+        // rather than panicking or looping indefinitely.
+        let lower = min;
+        let upper = max.saturating_add(1);
+        let partition_query = get_part_query(source_conn, &part.query, &part.column, lower, upper)?;
+        queries.push(partition_query);
+        return Ok(queries);
+    }
+
+    // Use checked arithmetic to avoid overflow when computing partition boundaries.
+    let range_len = max.checked_sub(min).and_then(|r| r.checked_add(1));
+    let partition_size = match range_len {
+        Some(len) => len / num,
+        None => {
+            // Overflow occurred in (max - min + 1); treat range as zero to avoid bad partitions.
+            throw!(anyhow!(
+                "partition range overflow: min={}, max={} is too large",
+                min,
+                max
+            ));
+        }
+    };
+
+    // partition_size == 0 means the range is smaller than the number of partitions.
+    // In this case we still create num partitions, but some will be empty.
+    let partition_size = partition_size.max(1);
 
     for i in 0..num {
-        let lower = min + i * partition_size;
-        let upper = match i == num - 1 {
-            true => max + 1,
-            false => min + (i + 1) * partition_size,
+        let lower = match min.checked_add((i as i64).saturating_mul(partition_size)) {
+            Some(l) => l,
+            None => {
+                // Overflow computing lower bound; clamp to min.
+                min
+            }
+        };
+        let upper = if i as i64 == num - 1 {
+            // Last partition goes up to and including max.
+            match max.checked_add(1) {
+                Some(u) => u,
+                None => max, // Saturating case; max is i64::MAX, last partition upper stays at max.
+            }
+        } else {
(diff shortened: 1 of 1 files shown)


connectorx/src/partition.rs
```diff
@@ -84,13 +84,60 @@ pub fn partition(part: &PartitionQuery, source_conn: &SourceConn) -> OutResult<V
...
 
     for i in 0..num {
-        let lower = min + i * partition_size;
-        let upper = match i == num - 1 {
-            true => max + 1,
-            false => min + (i + 1) * partition_size,
+        let lower = match min.checked_add((i as i64).saturating_mul(partition_size)) {
+            Some(l) => l,
+            None => {
+                // Overflow computing lower bound; clamp to min.
```
wangxiaoying, line 126: Won't this result in multiple partitions with overlapping ranges? This may cause the wrong result.

wangxiaoying (changes requested):
Thanks @hieuit095 for the PR!

I think the partition num check and max-min+1 overflow looks good. But the clamping logic may cause the same range handled by multiple partitions and result in the wrong result. We could just throw an error on those cases.

saurabh500:
@hieuit095 do you plan to carry this forward?

connectorx/src/partition.rs
saurabh500: It will. Good catch.

hieuit095:
Yes, definitely! Thanks for checking in, @saurabh500.

I've just rebased onto latest `main` (including the baseline unit tests from #956) and addressed the review feedback from @wangxiaoying:

1. **Removed clamping logic**: Replaced all `saturating_*` operations and clamping (`lower -> min`, `upper -> max`) with explicit checked arithmetic (`checked_mul`, `checked_add`). Any boundary calculation overflow now immediately returns an explicit error instead of silently overlapping partitions or duplicating data.
2. **Guarded upper bound overflow**: `max.checked_add(1)` is now checked to ensure that when `max == i64::MAX`, an overflow error is returned rather than wrapping or clamping.
3. **Updated & extended tests**:
   - Updated the division-by-zero test from expecting a panic to asserting a proper error is returned.
   - Added unit tests specifically verifying that integer overflows at range calculation, intermediate steps, and `i64::MAX` boundaries correctly return errors.
   - Updated the inverted range test to verify safe handling.

All 22 unit tests are passing locally with `cargo test --lib --features src_sqlite`. Ready for another look when you have time!

saurabh500:
The changes look good to me and the test changes make sense. The new changes look better than earlier too :) 

@wangxiaoying do you want to reset your vote?

connectorx/Cargo.toml
```diff
@@ -99,10 +99,12 @@ name = "connectorx"
 criterion = "0.3"
 env_logger = "0.11"
 iai = "0.1"
-pprof = { version = "0.14", features = ["flamegraph"] }
 testcontainers = { version = "0.23", features = ["blocking"] }
 ureq = { version = "2", features = ["json"] }
 
+[target.'cfg(not(target_os = "windows"))'.dev-dependencies]
```
saurabh500: Is this accidental? Why is this needed?

wangxiaoying:
> The changes look good to me and the test changes make sense. The new changes look better than earlier too :)
> 
> @wangxiaoying do you want to reset your vote?

I think the update looks good to me! Please feel free to merge this after the comment you left in the review is addressed.

connectorx/Cargo.toml
hieuit095: Good catch -- yes, that change was unrelated to this fix and shouldn't have been part of the PR. I've removed it so the diff is scoped back to the partitioning changes.

While revisiting the bounds logic, I also removed the remaining `partition_size.max(1)` fallback. When `num` is larger than the inclusive range size, that could produce partitions beyond the requested `max`, so that case now returns an explicit error instead. I added regression coverage for that case as well. Thanks for catching the unrelated change!

saurabh500 (commented):
Re-reviewed commit 6baee12bf079e03629bcbf6c144f1b5b0fdb64a9. The latest changes correctly reject zero partitions, inverted ranges, and arithmetic overflow while preserving existing partition counts and bounds for valid ranges, including the zero-sized partition behavior when num exceeds the range size. Focused partition tests pass with the src_sqlite feature enabled. No blocking issues found.

Issue #886, opened 2026-01-28
Oracle date range support: PanicException for dates outside [1677–2262] due to nanosecond timestamp limits


dbascoules:
Hi @wangxiaoying,

I’ve tested the latest version of connector-x with Oracle, and all existing tests pass successfully. However, I’m encountering an issue when extending the test data in oracle.sql to include two specific dates:

- One before 1677
- One after 2262

In this case, the test fails with the following error:

`FAILED connectorx/tests/test_oracle.py::test_oracle_types - pyo3_runtime.PanicException: out of range DateTime`

This is precisely the scenario I was trying to address in commit https://github.com/sfu-db/connector-x/pull/870/changes/b3a47b8d2c3eed52045a060681abf64d9efea42c by switching from timestamp[ns] (which supports ~585 years) to timestamp[us] (which supports ~290,000 years) to better accommodate Oracle’s wider date range.

Could you please advise on how best to handle this edge case? Should we aim to support such dates, or is it acceptable to document this limitation?

Thanks in advance for your guidance!

NicolasDortu:
This issue also impacts SQL Server, and I would guess any DBMS with timestamps after the year 2262. It is quite impactful, as it is not uncommon to have timestamps with the year 2999 or above to represent a sort of “infinite date”.

I'm not a Rust developer, but maybe having a parameter that would allow selecting ms instead of ns (which could be defaulting to ns so it doesn’t impact the current implementation) could be of great help.

Thanks a lot for the work you do with ConnectorX !

dbascoules:
Hi @wangxiaoying,

Just a friendly follow-up on this issue. We're currently planning our next data pipeline iterations and would love to align on how to handle these edge-case dates with connector-x.

A few quick questions:
1. Is supporting dates outside the [1677–2262] range something that could be considered for a future release?
2. Would a configurable option (e.g., `timestamp_mode="us"`) be an acceptable approach, as suggested by @NicolasDortu?
3. Or would you recommend documenting this as a known limitation for now?

Happy to help with testing or contributing a PR if that helps move this forward. Thanks again for your time and the great work on connector-x! 🙏

PR #901, opened 2026-04-13
Feat/dbascoules/default to microsecond precision

dbascoules:
## Summary
This PR adresses #886 by changing the default timestamp handling behavior in connector-x to use **microsecond precision (`timestamp[us]`)** instead of nanoseconds, unless the source data type explicitly supports nanosecond resolution.

### Real-world data sources rarely provide true nanosecond precision
- **Oracle**: `TIMESTAMP` type supports up to 9 digits, but in practice most deployments use 3-6 digits (milliseconds to microseconds).
- **PostgreSQL**: `timestamp` defaults to microsecond precision; nanosecond support requires explicit configuration and is uncommon in production.
- **SQL Server**: `datetime2` supports up to 7 digits of fractional seconds (~100ns), but not true nanoseconds.
- **MySQL**: `DATETIME(6)` and `TIMESTAMP(6)` cap at microsecond precision.
- **File formats**: Parquet, Arrow, and most interchange formats default to microsecond resolution for timestamps.

### Avoids silent overflows and `PanicException` errors
Using nanoseconds as the default forces all timestamps into a narrow valid range ([1677–2262]), causing runtime panics for legitimate historical or future dates. Microsecond precision extends the safe range to approximately **[290 BC – 2942 AD]**, covering virtually all real-world business data.

### Aligns with Arrow and pandas conventions
- PyArrow and pandas use `timestamp[us]` as their default temporal type for good reason: it balances precision, range, and interoperability.
- This change reduces friction when loading data into pandas DataFrames or Arrow tables without requiring manual casts.

### Preserves explicit nanosecond support when needed
When a source column explicitly declares nanosecond capability (e.g., with Oracle TimestampNano and TimestampTzNano), connector-x will still honor that and use `timestamp[ns]`.

## Impact
- **Backward-compatible for most users**: Microsecond precision is sufficient for >99% of analytical and operational workloads.
- **Fewer runtime errors**: Eliminates unexpected panics on edge-case dates.
- **Better interoperability**: Smoother integration with pandas, PyArrow, DuckDB, and other tools in the PyData ecosystem.
- **Potential breaking change**: Users relying on implicit nanosecond handling for sub-microsecond data may need to explicitly use compatible source types.

## Next Steps / Questions for Reviewers
1. Does this default align with the project's long-term vision for type inference?
2. Should we add a configuration flag (e.g., `timestamp_precision="auto" | "us" | "ns"`) to give users explicit control?
3. Are there specific connectors where nanosecond default is critical and should be exempted?

Happy to iterate on the implementation or add tests/docs as needed. Thanks for reviewing! 🙏


wangxiaoying:
Hi @dbascoules , thank you for the PR and the detailed explanation!

> Does this default align with the project's long-term vision for type inference?
I think it is OK to have microsecond as the default value, as long as the nanosecond support is preserved. Probably should mention this in the document and release note though.

> Should we add a configuration flag (e.g., timestamp_precision="auto" | "us" | "ns") to give users explicit control?
My understanding is that we could use the corresponding type (us or ns) according to the column type defined in the database for now as it won't loose any information. Users could convert the type of the result pandas/arrow column if it is needed.

> Are there specific connectors where nanosecond default is critical and should be exempted?
I cannot think of anything that should be exempted for now.

I think one additional thing should be done in this PR is to add the nanosecond type into the CI. For example, adding a column with `TIMESTAMP(9)` into the `test_types` table in `scripts/oracle.sql` and update the corresponding tests. We can merge this PR after it is done.

Let me know if you have further questions! Thanks again!

dbascoules:
Hi @wangxiaoying,

Thanks for the review and for catching the missing TIMESTAMP(9) test case — that's exactly the kind of coverage I should have included from the start, so I really appreciate you pointing it out!

I've now added:
* TIMESTAMP(9) and TIMESTAMP WITH TIME ZONE(9) columns to the test_types table in scripts/oracle.sql
* Updated the corresponding assertions to validate nanosecond precision handling when explicitly declared
* All CI checks are passing on my laptop (waiting for gihub CI)

Let me know if you'd like any further adjustments before merging. Happy to iterate!

Best,
David

wangxiaoying:
@dbascoules Thanks for the prompt response. I have merged the PR and I'm planning to release a version by the end of this month, with this modification included.

Thanks again for the contribution!

PR #902, opened 2026-04-19
Fix MySQL TINYINT incorrectly mapped to Boolean in Arrow transport

### LINKED ISSUE #494: MySQL TINYINT gets read as boolean into Arrow Dataframe

Python

The bug exists in the 0.3.1 and the current 0.3.2 dev version

MySQL

Arrow

Columns using a TINYINT(4) type are cast as a boolean when read into an Arrow dataframe in version 0.3.1 and onwards.
I don't know if this is happening across any other types. I have seen columns with large integers sometimes get cast as floats, though I don't have an example of this available.

```
CREATE TABLE `cx_test` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
   `test_column` tinyint(4) DEFAULT NULL,
   PRIMARY KEY (`id`)
 ) ENGINE=InnoDB AUTO_INCREMENT=0 DEFAULT CHARSET=latin1;

INSERT INTO `cx_test` (test_column)
  VALUES (0);
INSERT INTO `cx_test` (test_column)
  VALUES (1);
INSERT INTO `cx_test` (test_column)
  VALUES (2);
INSERT INTO `cx_test` (test_column)
  VALUES (3);
 ```

*With Version 0.3.0, everything works correctly:*
```
cx.read_sql(conn_str, "select * from cx_test", return_type="arrow")
pyarrow.Table
id: int64
test_column: int64
----
id: [[1,2,3,4]]
test_column: [[0,1,2,3]]
```

*With Version 0.3.1 and onwards, the data is converted into bools:*
```
cx.read_sql(conn_str, "select * from cx_test", return_type="arrow")
pyarrow.Table
id: int64
test_column: bool
----
id: [[1,2,3,4]]
test_column: [[false,true,true,true]]

### LINKED ISSUE #779: MySql tinyint gets converted to a boolean
This seems like a bug to me, but the MySQL `tinyint` type gets converted to a boolean value here:
https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/mysql_arrow.rs#L44

This change was made in https://github.com/sfu-db/connector-x/commit/2570c4c74a139566df301f0457f014663081cbf7

Looking at the [mysql documentation](https://dev.mysql.com/doc/refman/8.4/en/integer-types.html), a `tinyint` ranges from -128 to 127 for the signed version and 0 to 255 for the unsigned one. So it would seem like this should be mapped as a i8 at a minimum.

oystersuki:
## Summary

Fix a data corruption bug where MySQL `TINYINT` values outside {0, 1} are silently converted to boolean `true`, losing the original value.

Fixes #779
Fixes #494

## Problem

The MySQL → Arrow transport maps `Tiny[i8]` to `Boolean[bool]` with a conversion function `val != 0`. This means any non-zero TINYINT value (e.g. -128, 2, 3, 127) becomes `true`, silently destroying data.

MySQL `TINYINT` is commonly used for small integer values such as level codes (0-3) or status flags, not just booleans. This mapping was introduced in commit 2570c4c and was not present in v0.3.0.

## Fix

- Changed `Tiny[i8]` mapping from `Boolean[bool]` to `Int16[i16]` for both `BinaryProtocol` and `TextProtocol` in both Arrow and ArrowStream transports
- Removed the manual `TypeConversion<i8, bool>` implementation from both transports

`Int16` is used because `ArrowTypeSystem` does not currently include an `Int8` variant (PostgreSQL has no 8-bit integer type, so #743 did not add one). `Int16` can represent the full signed TINYINT range (-128 to 127) without loss.

## Consistency

The MySQL → **pandas** transport in this same codebase already maps `Tiny[i8]` to `I64[i64]`, and the existing test `test_mysql_types_binary` in the Python test suite verifies that TINYINT values like -128 and 127 round-trip correctly as integers. This PR brings the Arrow and ArrowStream transports in line with that established behavior.

## Tests

Added regression tests (`test_mysql_tinyint_not_bool` and `test_mysql_tinyint_not_bool_text`) that verify TINYINT values -128 and 127 are preserved as `Int16` through both protocols.

All 10 MySQL Arrow tests pass locally.

### DIFF
--- a/connectorx/src/transports/mysql_arrow.rs
+++ b/connectorx/src/transports/mysql_arrow.rs
@@ -41,7 +41,7 @@ impl_transport!(
     mappings = {
         { Float[f32]                 => Float64[f64]            | conversion auto }
         { Double[f64]                => Float64[f64]            | conversion auto }
-        { Tiny[i8]                   => Boolean[bool]           | conversion option }
+        { Tiny[i8]                   => Int16[i16]              | conversion auto }
         { Short[i16]                 => Int64[i64]              | conversion auto }
         { Int24[i32]                 => Int64[i64]              | conversion none }
         { Long[i32]                  => Int64[i64]              | conversion auto }
@@ -77,7 +77,7 @@ impl_transport!(
     mappings = {
         { Float[f32]                 => Float64[f64]            | conversion auto }
         { Double[f64]                => Float64[f64]            | conversion auto }
-        { Tiny[i8]                   => Boolean[bool]           | conversion option }
+        { Tiny[i8]                   => Int16[i16]              | conversion auto }
         { Short[i16]                 => Int64[i64]              | conversion auto }
         { Int24[i32]                 => Int64[i64]              | conversion none }
         { Long[i32]                  => Int64[i64]              | conversion auto }
@@ -127,11 +127,5 @@ impl<P> TypeConversion<Decimal, f64> for MySQLArrowTransport<P> {
 impl<P> TypeConversion<Value, String> for MySQLArrowTransport<P> {
     fn convert(val: Value) -> String {
         to_string(&val).unwrap()
-    }
-}
-
-impl<P> TypeConversion<i8, bool> for MySQLArrowTransport<P> {
-    fn convert(val: i8) -> bool {
-        val != 0
     }
 }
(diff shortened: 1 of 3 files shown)


wangxiaoying:
Thank you @oystersuki for the PR! The code looks good and has been merged.

brainbytes42:
Thanks for fixing the issue, I ran into it as well.

The last Release is from January, this Fix from April, now it's September - when will the next release come so that this Fix will get actually deployed? 
(I don't want to be rude, but wandering around, I found several Fixes for different Problems all relating to wrong data produced - but no release publishing them yet... Issue #494, that was fixed as well as #779 is from 2023...)

wangxiaoying:
Hi @brainbytes42 , thanks for the comment! We have a new alpha release recently (`pip install connectorx==0.4.6a2`) that has this and many earlier fixes included, please feel free to try it out. We will have a stable release after verifying this alpha release.

Issue #890, opened 2026-02-25
Singlestore BIGINT UNSIGNED columns returned incorrectly


dron-dronych:
#### What language are you using?

**Python 3.10**.

#### What version are you using?

connectorx 0.4.5

#### What database are you using?

Singlestore 5.7.32

#### What dataframe are you using?

Pandas

#### Can you describe your bug?

Having a BIGINT UNSIGNED column with big values are returned incorrectly. I have noticed the last four digits are different from the actual values.

##### Example query / code

```
create table test
(id bigint unsigned not null);

insert into test
(id) values
(10), (20), (1), (2087271583674132669);

df = cx.read_sql('singlestoredb://root:root@localhost:3306/testdb', 'select id from test;')

df['id'].astype(int)
```

#### What is the error?

2087271583674132736 instead of 2087271583674132669

dron-dronych:
after some experiments i found the native connection via their dedicated client and through pandas.read_sql() gives the proper uint64 (expected) so maybe there is an issue with connectorx. probably the admins should reopen this @wangxiaoying

PR #903, opened 2026-04-30
Fix MySQL BIGINT UNSIGNED mapped to Float64 causing precision loss in Arrow transport

### LINKED ISSUE #664: MySQL ULongLong should be treated as integer

Python

0.3.3

MySQL

Arrow

MySQL ULongLong is treated as Float in conversion instead of LongLong.

This seem to be coming from here:
https://github.com/sfu-db/connector-x/blob/main/connectorx/src/transports/mysql_arrow.rs#L48-L53

```
        { LongLong[i64]              => Int64[i64]              | conversion auto }
        { UTiny[u8]                  => Int64[i64]              | conversion auto }
        { UShort[u16]                => Int64[i64]              | conversion auto }
        { ULong[u32]                 => Int64[i64]              | conversion auto }
        { UInt24[u32]                => Int64[i64]              | conversion none }
        { ULongLong[u64]             => Float64[f64]            | conversion auto }  <--- 
```

oystersuki:
## Summary

Fix incorrect type mapping where MySQL `BIGINT UNSIGNED` is converted to Float64 instead of UInt64 in the Arrow transport. This causes precision loss for values exceeding 2^53.

Fixes #664

## Problem

The MySQL → Arrow transport maps `ULongLong[u64]` to `Float64[f64]`. IEEE 754 Float64 can only represent integers exactly up to 2^53 (9007199254740992). Values above this threshold silently lose precision.

As #664 points out, `LongLong[i64]` is already correctly mapped to `Int64[i64]`, while `ULongLong[u64]` was the only unsigned integer type mapped to Float64.

## Fix

Changed `ULongLong[u64]` mapping from `Float64[f64]` to `UInt64[u64]` for both `BinaryProtocol` and `TextProtocol` in both Arrow and ArrowStream transports.

`ArrowTypeSystem` already includes a `UInt64` variant with full support in `arrow_assoc.rs`, so no destination-layer changes are needed.

## Scope

This PR addresses MySQL Arrow and ArrowStream transports only.

The pandas transport also maps `ULongLong` to `F64`, causing the same precision loss (as reported in #890). Fixing the pandas side would require adding a `UInt64` variant to `PandasTypeSystem`, which is a larger change left for a separate PR.

## Consistency

I checked all Arrow transports in this codebase. MySQL was the only source where unsigned 64-bit integers were mapped to Float64. The ClickHouse Arrow transport already correctly maps `UInt64[u64] → UInt64[u64]`. No other source database (PostgreSQL, MSSQL, Oracle, SQLite, BigQuery, Trino) has unsigned 64-bit integer types.

Other MySQL unsigned integer types are also mapped to integer Arrow types (`UTiny → Int64`, `UShort → Int64`, `ULong → Int64`), making `ULongLong` the only unsigned type that was converted to Float64.

## Impact

This fix also applies to databases using the MySQL wire protocol, such as SingleStore, when accessed via Arrow or ArrowStream return types.

## Tests

Added regression tests (`test_mysql_bigint_unsigned_not_float` and `test_mysql_bigint_unsigned_not_float_text`) that verify BIGINT UNSIGNED values are returned as `UInt64` through both protocols.

### DIFF
--- a/connectorx/src/transports/mysql_arrow.rs
+++ b/connectorx/src/transports/mysql_arrow.rs
@@ -50,7 +50,7 @@ impl_transport!(
         { UShort[u16]                => Int64[i64]              | conversion auto }
         { ULong[u32]                 => Int64[i64]              | conversion auto }
         { UInt24[u32]                => Int64[i64]              | conversion none }
-        { ULongLong[u64]             => Float64[f64]            | conversion auto }
+        { ULongLong[u64]             => UInt64[u64]             | conversion auto }
         { Date[NaiveDate]            => Date32[NaiveDate]       | conversion auto }
         { Time[NaiveTime]            => Time64Micro[NaiveTimeWrapperMicro]       | conversion option }
         { Datetime[NaiveDateTime]    => Date64Micro[NaiveDateTimeWrapperMicro]   | conversion option }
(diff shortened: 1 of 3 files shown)


wangxiaoying:
Thanks @oystersuki for the PR!

PR #904, opened 2026-05-07
[codex] Reduce ODBC batch fetch copying

### LINKED ISSUE #15: Output None for null string

tomanizer:
## What changed

- Replaced the ODBC row-major `Vec<Option<OdbcCell>>` parser buffer with a reusable columnar batch snapshot.
- Added lightweight `OdbcValue` views so primitive values are read from compact column vectors and byte/string values are borrowed from the batch snapshot until the destination requests ownership.
- Reused per-column vectors and text/binary byte slots across batch fetches to reduce allocation churn.
- Updated the generic ODBC Arrow fallback to use the new value view and avoid extra owned byte clones for direct `AnySlice` text/binary reads.
- Added a focused unit test covering borrowed byte views from the columnar batch.

## Why

Issue #15 identified that ODBC-backed sources flattened every fetched cell into owned row-buffer cells before Arrow writes. That created extra copying and allocation pressure for large scans, wide tables, and text/binary-heavy workloads.

The parser still cannot hold borrowed `ColumnarAnyBuffer` slices directly across `fetch_next`, because the fetched batch borrows the cursor. This change makes that boundary explicit: the parser snapshots each batch in columnar form, reuses those buffers, and only clones into `String`/`Vec<u8>` when the `Produce` destination requires ownership.

Closes #15.

## Validation

- `cargo fmt --all --check`
- `cargo check -p connectorx --no-default-features --features "fptr src_odbc dst_arrow"`
- `cargo check -p connectorx --no-default-features --features "fptr src_db2 dst_arrow"`
- `cargo check -p connectorx --no-default-features --features "fptr src_sybase dst_arrow"`
- `cargo test -p connectorx --lib --no-default-features --features "fptr src_odbc" sources::odbc_core::tests`
- `cargo bench -p connectorx --no-default-features --features "fptr src_odbc dst_arrow" --bench odbc --no-run`


tomanizer:
Closing because this was opened against upstream too early. We will iterate in the fork first.

PR #910, opened 2026-05-14
fix(arrow_stream): propagate errors instead of panicking on connection setup

### LINKED ISSUE #909: arrow_stream + sslmode=require panics with PostgresPoolError(Error(None)) — works in arrow mode

Python

connectorx 0.4.4

PostgreSQL (AWS RDS, `sslmode=require`)

Arrow / Arrow Streaming

`return_type="arrow_stream"` panics at `connectorx/src/get_arrow.rs:318:22` (or `:355` with `protocol="cursor"`) during the initial Postgres connection setup when the URL has `sslmode=require`. The wrapped error is stripped to `PostgresPoolError(Error(None))` because the streaming path uses `.unwrap()` where the bulk path (`get_arrow`) uses `?`.

The exact same URL works in `return_type="arrow"`.

```python
import connectorx as cx

url = "postgresql://user:pass@host:5432/db?sslmode=require"

# Works
arrow = cx.read_sql(url, "SELECT 1", return_type="arrow")
print("arrow OK:", arrow.num_rows)

# Panics
reader = cx.read_sql(url, "SELECT 1", return_type="arrow_stream")
list(reader)
```

#### What is the error?

```
thread '<unnamed>' panicked at /Users/.../connectorx/src/get_arrow.rs:318:22:
called `Result::unwrap()` on an `Err` value: PostgresPoolError(Error(None))

(r2d2 runtime log surfaces the real cause:)
[r2d2] error communicating with the server: Connection reset by peer (os error 54)
```

#### Test matrix

|                                 | No SSL | sslmode=require |
| ------------------------------- | :----: | :-------------: |
| `return_type="arrow"`           |   ✓    |        ✓        |
| `return_type="arrow_stream"`    |   ✓    |   ✗ panic       |

Tried `protocol={binary, cursor, simple}` — `binary` and `cursor` both panic identically at different line numbers (`get_arrow.rs:318` and `:355`); `simple` is not implemented. Connection environment doesn't matter (reproducible with direct DSN and over an SSH tunnel).

#### Related

- #864 (closed) — different typesystem mappings between `arrow` and `arrow_stream`
- #866 (open) — timestamp columns wrong type in `arrow_stream`
- #511 (open) — connectorx + SSH tunnel hang (different failure mode)

These all point at `arrow_stream` being a separately-implemented codepath with rougher edges than `arrow`. This issue adds SSL handshake to that list.

#### Suggested fix (visibility, at minimum)

Two `.unwrap()` calls in `new_record_batch_iter` should be `?` (with `#[throws(...)]` on the function) to mirror `get_arrow`'s error handling. That alone would surface the underlying r2d2 error instead of stripping it to `Error(None)`. The root cause (why r2d2 connection setup fails under `arrow_stream` but not `arrow` with identical inputs) is a separate question worth tracing.

ricardo-valero:
## Summary

Partially addresses #909 — `return_type=\"arrow_stream\"` panicked at `connectorx/src/get_arrow.rs:318:22` (or `:355` for cursor) with `PostgresPoolError(Error(None))` whenever the initial Postgres connection setup failed (notably with `sslmode=require`), while the equivalent `return_type=\"arrow\"` succeeded.

The root cause is that `new_record_batch_iter` used `.unwrap()` everywhere, panicking the thread and masking the wrapped r2d2 error. `get_arrow` already returned errors via `#[throws(ConnectorXOutError)]` and `?`; this change brings the streaming path in line.

### Changes

- `connectorx/src/get_arrow.rs` — `new_record_batch_iter` is now `#[throws(ConnectorXOutError)]`; every `.unwrap()` becomes `?`. Returns are typed through `let iter: Box<dyn RecordBatchIterator>` so the unsizing coercion is explicit.
- `connectorx/src/errors.rs` — adds `From<…>` impls on `ConnectorXOutError` for each `*ArrowStream*` transport error so `?` can wrap them.
- `connectorx-python/src/arrow.rs` — propagates the new `Result` with `?` (the surrounding function already returns `ConnectorXPythonError`, which has a `From<ConnectorXOutError>`).
- `connectorx-cpp/src/lib.rs` — keeps current behavior with an explicit `.unwrap()` (FFI boundary; same as the surrounding code).
- `connectorx-python/connectorx/tests/test_postgres.py` — adds a TLS arrow_stream regression test and an explicit “bad host raises” test that exercises the error path.

## Test plan

- [x] `cargo check` + `cargo build --all-features` on the workspace
- [x] `cargo test --lib` (no unit tests defined for this module)
- [ ] CI integration tests against a TLS-enabled Postgres (`POSTGRES_URL_TLS`) — `test_postgres_tls_arrow_stream` and `test_postgres_tls_arrow_stream_bad_host_raises`
- [ ] Manual repro of the issue against AWS RDS with `sslmode=require` to confirm the panic is replaced with a `RuntimeError` carrying the underlying r2d2 message

### DIFF
--- a/connectorx-cpp/src/lib.rs
+++ b/connectorx-cpp/src/lib.rs
@@ -281,7 +281,8 @@ pub unsafe extern "C" fn connectorx_scan_iter(
     }
 
     let arrow_iter: Box<dyn RecordBatchIterator> =
-        new_record_batch_iter(&source_conn, None, query_vec.as_slice(), batch_size, None);
+        new_record_batch_iter(&source_conn, None, query_vec.as_slice(), batch_size, None)
+            .unwrap();
 
     Box::into_raw(Box::new(arrow_iter))
 }
--- a/connectorx-python/connectorx/tests/test_postgres.py
+++ b/connectorx-python/connectorx/tests/test_postgres.py
@@ -888,6 +888,33 @@ def test_postgres_tls_disable(postgres_url_tls: str) -> None:
     not os.environ.get("POSTGRES_URL_TLS"),
     reason="Do not test Postgres TLS unless `POSTGRES_URL_TLS` is set",
 )
+def test_postgres_tls_arrow_stream(postgres_url_tls: str) -> None:
+    query = "SELECT * FROM test_table"
+    reader = read_sql(
+        f"{postgres_url_tls}?sslmode=require",
+        query,
+        return_type="arrow_stream",
+    )
+    batches = list(reader)
+    assert len(batches) > 0
+    assert batches[0].num_columns > 0
+
+
+@pytest.mark.skipif(
+    not os.environ.get("POSTGRES_URL_TLS"),
+    reason="Do not test Postgres TLS unless `POSTGRES_URL_TLS` is set",
+)
+def test_postgres_tls_arrow_stream_bad_host_raises(postgres_url_tls: str) -> None:
+    bad_url = "postgresql://user:pass@127.0.0.1:1?sslmode=require"
+    with pytest.raises(RuntimeError):
+        reader = read_sql(bad_url, "SELECT 1", return_type="arrow_stream")
+        list(reader)
+
+
+@pytest.mark.skipif(
+    not os.environ.get("POSTGRES_URL_TLS"),
+    reason="Do not test Postgres TLS unless `POSTGRES_URL_TLS` is set",
+)
 @pytest.mark.xfail
 def test_postgres_tls_fail(postgres_url_tls: str) -> None:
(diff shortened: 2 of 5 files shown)


oystersuki:
Thank you @ricardo-valero ! I pushed a few commits to your branch: merged `main`, ran `cargo fmt --all` (the CI formatting check was failing), and replaced the bad-host test with one that fails on an invalid `sslmode` so it runs in CI without TLS and doesn't wait 30s for the connection timeout.

I also changed "Fixes #909" to "Partially addresses #909", since the `sslmode=require` half of that issue isn't covered here.

PR #915, opened 2026-05-31
fix(mysql): detect binary vs text columns by charset (charsetnr == 63) instead of BINARY_FLAG

oystersuki:
## Summary

MySQL `CHAR`/`VARCHAR`/`TEXT` columns that use a `*_bin` collation (e.g. `ascii_bin`, `utf8mb4_bin`) are incorrectly returned as binary (Arrow `large_binary`) instead of strings.

The binary-vs-text decision in the MySQL source relied on `ColumnFlags::BINARY_FLAG`, but that flag is set for any column using a binary (`_bin`) collation — which are still **text** columns. This PR switches the decision to the only reliable signal, the **binary charset (collation id 63)**, and removes a related, fragile `charset == utf8mb4_general_ci` special case.

## Reproduction

```sql
CREATE TABLE cx_typetest (
  vc_ci    VARCHAR(20) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci,
  vc_bin   VARCHAR(20) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin,
  vc_ascii VARCHAR(20) CHARACTER SET ascii   COLLATE ascii_bin,
  txt_ci   TEXT        CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci,
  txt_bin  TEXT        CHARACTER SET utf8mb4 COLLATE utf8mb4_bin,
  vb       VARBINARY(20),
  bl       BLOB
);
```

Arrow schema from `cx.read_sql(url, "SELECT * FROM cx_typetest", return_type="arrow")` (MySQL 8.0, utf8mb4 connection):

| column | definition | 0.4.5 | 0.4.6-alpha.1 | this PR |
|---|---|---|---|---|
| vc_ci | VARCHAR utf8mb4_general_ci | string | string | string |
| vc_bin | VARCHAR utf8mb4_bin | string | **large_binary** ❌ | string ✅ |
| vc_ascii | VARCHAR ascii_bin | string | **large_binary** ❌ | string ✅ |
| txt_ci | TEXT utf8mb4_general_ci | string | string | string |
| txt_bin | TEXT utf8mb4_bin | string | string | string |
| vb | VARBINARY | **string** ❌ | large_binary | large_binary |
| bl | BLOB | large_binary | large_binary | large_binary |

- **0.4.5**: `VARBINARY` was wrongly returned as `string`.
- **0.4.6-alpha.1** (fixed `BINARY`/`VARBINARY` in e708b11): correct for `vb`, but introduced a regression — `CHAR`/`VARCHAR` with a `*_bin` collation are now returned as binary.
- **This PR**: `*_bin` text columns are strings again, while `VARBINARY`/`BLOB` stay binary. It is a strict superset of the e708b11 fix.

## Root cause

1. e708b11 distinguished `BINARY`/`VARBINARY` from `CHAR`/`VARCHAR` (which share the `MYSQL_TYPE_STRING` / `MYSQL_TYPE_VAR_STRING` type codes) using `ColumnFlags::BINARY_FLAG`. But `BINARY_FLAG` indicates the column uses a binary (`_bin`) **collation**, not that it stores binary data — `ascii_bin` / `utf8mb4_bin` columns set the flag yet hold text.
2. `fetch_metadata` separately treated `*_BLOB` columns as text only when `charset == UTF8MB4_GENERAL_CI`. The result `charsetnr` depends on the negotiated connection collation (e.g. `utf8mb4_0900_ai_ci` = 255 on MySQL 8.0 vs `utf8mb4_general_ci` = 45), so under other connection collations TEXT columns could also fall through to binary.

## Why `charsetnr == 63`

MySQL documents charset id 63 (the `binary` charset) as the canonical way to tell binary data from text in result metadata. The [C API Developer Guide](https://dev.mysql.com/doc/c-api/8.0/en/c-api-prepared-statement-type-conversions.html) specifies that a `charsetnr` of 63 is what distinguishes `BINARY` from `CHAR`, `VARBINARY` from `VARCHAR`, and the `BLOB` types from the `TEXT` types (see also [C API Basic Data Structures](https://dev.mysql.com/doc/c-api/8.0/en/c-api-data-structures.html) for `MYSQL_FIELD.charsetnr`).

Per [Reference Manual §13.3.3](https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html), `CHAR BINARY` / `VARCHAR BINARY` (i.e. a `_bin` collation) still store *nonbinary character strings* — they are text, distinct from the `BINARY`/`VARBINARY` data types. So `BINARY_FLAG` (the `_bin` attribute) is the wrong discriminator; the binary charset (63) is the right one. The same mistake has surfaced in other Rust MySQL drivers (e.g. [sqlx#3387](https://github.com/launchbadge/sqlx/issues/3387)).

## Changes

- `sources/mysql/typesystem.rs`: thread the column charset into `MySQLTypeSystem::from`; `MYSQL_TYPE_STRING` / `MYSQL_TYPE_VAR_STRING` and the four `*_BLOB` (TEXT/BLOB) arms now map to a binary type iff `charsetnr == 63`, otherwise to text.
- `sources/mysql/mod.rs`: pass `col.character_set()` at both `fetch_metadata` call sites; remove the now-redundant `charset == UTF8MB4_GENERAL_CI` special case (and the unused server-version check).
- `partition.rs`: update the `mysql_get_partition_range` call site.

## Tests

- Unit tests (no DB) for the type mapping: `BINARY`/`VARBINARY`/`BLOB` (charset 63) → binary; `CHAR`/`VARCHAR`/`TEXT` → text; and `*_bin` text columns → text despite `BINARY_FLAG` being set.
- Integration test against a new `test_str_collation` table asserting the Arrow schema (string vs binary) for both the binary and text protocols.

### Updated test_mediumtext expectations from bytes to str

MEDIUMTEXT is a text type and was previously returned as binary because the charset == utf8mb4_general_ci special case did not match the CI connection collation (utf8mb4_0900_ai_ci). This is exactly the fragility this PR removes.

### DIFF
--- a/connectorx-python/connectorx/tests/test_mysql.py
+++ b/connectorx-python/connectorx/tests/test_mysql.py
@@ -336,7 +336,7 @@ def test_mysql_types_binary(mysql_url: str) -> None:
                 dtype="object",
             ),
             "test_mediumtext": pd.Series(
-                [None, b"", b"medium text!!!!"], dtype="object"
+                [None, "", "medium text!!!!"], dtype="object"
             ),
             "test_bit": pd.Series(
                 [b'\x17', b'\x18', None]
@@ -416,7 +416,7 @@ def test_mysql_types_text(mysql_url: str) -> None:
                 dtype="object",
             ),
             "test_mediumtext": pd.Series(
-                [None, b"", b"medium text!!!!"], dtype="object"
+                [None, "", "medium text!!!!"], dtype="object"
             ),
             "test_bit": pd.Series(
                 [b'\x17', b'\x18', None]
--- a/connectorx/src/partition.rs
+++ b/connectorx/src/partition.rs
@@ -282,8 +282,11 @@ fn mysql_get_partition_range(conn: &Url, query: &str, col: &str) -> (i64, i64) {
         .query_first(range_query)?
         .ok_or_else(|| anyhow!("mysql range: no row returns"))?;
 
-    let col_type =
-        MySQLTypeSystem::from((&row.columns()[0].column_type(), &row.columns()[0].flags()));
+    let col_type = MySQLTypeSystem::from((
+        &row.columns()[0].column_type(),
+        &row.columns()[0].flags(),
+        row.columns()[0].character_set(),
+    ));
 
     let (min_v, max_v) = match col_type {
         MySQLTypeSystem::Tiny(_) => {
--- a/connectorx/src/sources/mysql/mod.rs
+++ b/connectorx/src/sources/mysql/mod.rs
@@ -17,14 +17,7 @@ use fehler::{throw, throws};
 use log::{debug, warn};
 use r2d2::{Pool, PooledConnection};
 use r2d2_mysql::{
-    mysql::{
-        consts::{
-            ColumnFlags as MySQLColumnFlags, ColumnType as MySQLColumnType, UTF8MB4_GENERAL_CI,
-            UTF8_GENERAL_CI,
-        },
-        prelude::Queryable,
-        Binary, Opts, OptsBuilder, QueryResult, Row, Text,
-    },
+    mysql::{prelude::Queryable, Binary, Opts, OptsBuilder, QueryResult, Row, Text},
     MySqlConnectionManager,
 };
 use rust_decimal::Decimal;
@@ -109,8 +102,6 @@ where
         assert!(!self.queries.is_empty());
 
         let mut conn = self.pool.get()?;
-        let server_version_post_5_5_3 = conn.server_version() >= (5, 5, 3);
-
         let first_query = &self.queries[0];
 
         match conn.prep(first_query) {
@@ -120,26 +111,11 @@ where
                     .iter()
                     .map(|col| {
                         let col_name = col.name_str().to_string();
-                        let col_type = col.column_type();
-                        let col_flags = col.flags();
-                        let charset = col.character_set();
-                        let charset_is_utf8 = (server_version_post_5_5_3
-                            && charset == UTF8MB4_GENERAL_CI)
-                            || (!server_version_post_5_5_3 && charset == UTF8_GENERAL_CI);
-                        if charset_is_utf8
-                            && (col_type == MySQLColumnType::MYSQL_TYPE_LONG_BLOB
-                                || col_type == MySQLColumnType::MYSQL_TYPE_BLOB
-                                || col_type == MySQLColumnType::MYSQL_TYPE_MEDIUM_BLOB
-                                || col_type == MySQLColumnType::MYSQL_TYPE_TINY_BLOB)
-                        {
-                            return (
-                                col_name,
-                                MySQLTypeSystem::Char(
-                                    !col_flags.contains(MySQLColumnFlags::NOT_NULL_FLAG),
-                                ),
-                            );
-                        }
-                        let d = MySQLTypeSystem::from((&col_type, &col_flags));
+                        let d = MySQLTypeSystem::from((
+                            &col.column_type(),
+                            &col.flags(),
+                            col.character_set(),
+                        ));
                         (col_name, d)
                     })
                     .unzip();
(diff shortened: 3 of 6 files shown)


wangxiaoying:
Thank you @oystersuki for the contribution! The PR looks good to me.

P.S. I’ve sent you a collaborator invite for the project. Once you accept it, you should have more access to the project.

oystersuki:
@wangxiaoying 
Thanks for the merge and collaborator invite!
Could you share how you'd like maintenance to be handled going forward? I'm wondering how much I should take on with reviews and merges, and whether there are areas you'd prefer to keep for yourself.

wangxiaoying:
Hi @oystersuki , thanks for accepting the invitation and willing to contribute to the project!

Feel free to handle reviews and merge PRs as you see fit. I’m happy for you to take an active role on maintenance in general (I currently may not always have the capacity for day-to-day maintenance, there’s no need to wait on me for routine reviews or merges).

For larger or more significant changes, feel free to ping me if you’d like my feedback or want to discuss anything. And if you think it would be a good point to cut a new alpha/stable release, just let me know and I’m happy to handle the release side.

oystersuki:
@wangxiaoying 

Thanks, that's very helpful to know.

On that note, there's one thing I'd like your input on. Since 0.4.6-alpha, several changes have landed that alter the output Arrow types on the MySQL side: TINYINT(1) no longer mapping to boolean, BIGINT UNSIGNED no longer widening to float, VARCHAR with binary collation, integer/float width narrowing, and timestamp precision(not only mysql). Individually these are all fixes, but together they're a fairly visible change for downstream users — Polars uses ConnectorX as the default engine for read_database_uri, so the schema of the resulting DataFrame will change for MySQL sources.

Could you cut an alpha2? I'll validate it against real MySQL workloads (via dlt, in a production pipeline) for a week or two before suggesting a stable release.

oystersuki:
@wangxiaoying

Thanks, that's very helpful to know.

On that note, there's one thing I'd like your input on. Since 0.4.6-alpha, several changes have landed that alter the output Arrow types on the MySQL side: TINYINT(1) no longer mapping to boolean, BIGINT UNSIGNED no longer widening to float, VARCHAR with binary collation, integer/float width narrowing, and timestamp precision(not only mysql). Individually these are all fixes, but together they're a fairly visible change for downstream users — Polars uses ConnectorX as the default engine for read_database_uri, so the schema of the resulting DataFrame will change for MySQL sources.

Could you cut an alpha2? I'll validate it against MySQL workloads (via dlt) for a week or two before suggesting a stable release.

wangxiaoying:
> @wangxiaoying
> 
> Thanks, that's very helpful to know.
> 
> On that note, there's one thing I'd like your input on. Since 0.4.6-alpha, several changes have landed that alter the output Arrow types on the MySQL side: TINYINT(1) no longer mapping to boolean, BIGINT UNSIGNED no longer widening to float, VARCHAR with binary collation, integer/float width narrowing, and timestamp precision(not only mysql). Individually these are all fixes, but together they're a fairly visible change for downstream users — Polars uses ConnectorX as the default engine for read_database_uri, so the schema of the resulting DataFrame will change for MySQL sources.
> 
> Could you cut an alpha2? I'll validate it against MySQL workloads (via dlt) for a week or two before suggesting a stable release.

Sure, I will release a new alpha version this week and let you know once it's done.

wangxiaoying:
Hi @oystersuki , 0.4.6a2 is available now!

oystersuki:
Hi @wangxiaoying

I'll verify it and let you know the results briefly. Thank you release!!

oystersuki:
Hi @wangxiaoying ! Not sure if this is the right place to contact you, but dropping this here.

I've been running 0.4.6a2 on MySQL 5~8 for a bit now, and it works fine on my end. I can't speak for other data sources, but it feels ready for a full release.

After the official release, I'm planning to work on improving the error types.

wangxiaoying:
> Hi @wangxiaoying ! Not sure if this is the right place to contact you, but dropping this here.
> 
> I've been running 0.4.6a2 on MySQL 5~8 for a bit now, and it works fine on my end. I can't speak for other data sources, but it feels ready for a full release.
> 
> After the official release, I'm planning to work on improving the error types.

Hi @oystersuki , thanks so much for testing! I think we can have a stable release, since we haven't heard any issue with the alpha release for now, and we probably want to have one before supporting the new python version (3.15) coming soon anyway. I will do it by this week.

PR #905, opened 2026-05-10
preserve integer dtype sizes instead of upcasting to int64

lukgra:
## Summary

MySQL integer types smaller than BIGINT were being unnecessarily upcast 
to larger types in the Arrow transport layer, wasting memory.

Closes #688

## Changes
Updated type mappings in `mysql_arrow.rs` for both `BinaryProtocol` and 
`TextProtocol`:

- `TINYINT` (i8)   → `Int16` (smallest available, no Int8 in Arrow)
- `SMALLINT` (i16) → `Int16` (was Int64)
- `MEDIUMINT` (i32)→ `Int32` (was Int64)
- `INT` (i32)      → `Int32` (was Int64)
- `BIGINT` (i64)   → `Int64` (unchanged)
- `TINYINT UNSIGNED` (u8)  → `UInt16` (no UInt8 in Arrow)
- `SMALLINT UNSIGNED` (u16)→ `UInt16` (was Int64)
- `MEDIUMINT UNSIGNED`(u32)→ `UInt32` (was Int64)
- `INT UNSIGNED` (u32)     → `UInt32` (was Int64)
- `BIGINT UNSIGNED` (u64)  → `UInt64` (was Float64, lossy!)
- `YEAR` (i16)             → `Int16` (was Int64)

Updated `test_mysql.rs` to reflect the corrected types.

## Notes
`ULongLong → UInt64` impactful fix (previous mapping  to `Float64` was lossy for large unsigned 64-bit values)

PR #919, opened 2026-06-22
feat(mysql): preserve narrow integer and float types in Arrow transports

oystersuki:
## Summary

MySQL's Arrow transports widen several column types more than necessary. This PR maps them to their narrowest faithful Arrow type, continuing the work started in #905 (which addressed only the `arrow` destination and stopped at an `Int16`/`UInt16` floor).

## What changed

**Type systems (`arrow` and `arrowstream` destinations)**

Added `Int8` / `UInt8` variants to `ArrowTypeSystem` and the matching `ArrowAssoc` impls (`Int8Builder` / `UInt8Builder`).

**`mysql_arrow` transport**

| MySQL | Before | After |
|-------|--------|-------|
| `FLOAT` | `Float64` | `Float32` |
| `TINYINT` | `Int16` | `Int8` |
| `TINYINT UNSIGNED` | `UInt16` | `UInt8` |

**`mysql_arrowstream` transport**

The streaming transport previously upcast almost every integer to `Int64` and mapped *unsigned* types to *signed* `Int64`. It now preserves width and signedness, matching the `arrow` destination:

| MySQL | Before | After |
|-------|--------|-------|
| `FLOAT` | `Float64` | `Float32` |
| `TINYINT` | `Int16` | `Int8` |
| `SMALLINT` | `Int64` | `Int16` |
| `MEDIUMINT` / `INT` | `Int64` | `Int32` |
| `TINYINT UNSIGNED` | `Int64` (signed!) | `UInt8` |
| `SMALLINT UNSIGNED` | `Int64` (signed!) | `UInt16` |
| `MEDIUMINT/INT UNSIGNED` | `Int64` (signed!) | `UInt32` |
| `YEAR` | `Int64` | `Int16` |

`BIGINT`/`BIGINT UNSIGNED` (`Int64`/`UInt64`) and the temporal, decimal, string and blob mappings are unchanged.

## Compatibility / behavior change

This changes the **output Arrow schema**: affected columns now have a narrower width, and (for the streaming transport) unsigned columns are now returned as unsigned types instead of signed `Int64`. No data is lost or truncated — every value fits in the narrower type, and `BIGINT UNSIGNED` remains `UInt64` (it is never demoted to `Float64`). Downstream consumers that relied on the previous, wider/signed schema may need to adjust.

## Relation to prior work

- #688 — "Int dtypes are unnecessarily big"
- #905 — preserved sizes for the `arrow` destination down to `Int16`

This PR lowers the floor to `Int8`/`UInt8`, removes the unnecessary `FLOAT -> Float64` promotion, and brings the `arrowstream` destination up to parity (it was not touched by #905).

## Testing

Regression tests were added to `connectorx/tests/test_mysql.rs` for both the binary and text protocols, covering `YEAR`, `FLOAT`, and the full set of signed/unsigned integer widths. Boundary values already present in the `test_types` seed are used (e.g. `YEAR = 2155`, which does not fit in `Int8`, and `TINYINT UNSIGNED = 255`, which does not fit in `Int8`), so the tests fail immediately on any accidental narrowing or sign error. No seed changes were required.

## Why there is no `arrow_stream` test

The integration test file `tests/test_mysql.rs` is compiled only under the `dst_arrow` feature; there is currently no `dst_arrowstream` integration test. The `arrow` and `arrowstream` destinations duplicate their entire surface (separate type systems, transports and `ArrowAssoc` impls), so a test written against one does not exercise the other.

Adding proper `arrow_stream` coverage cleanly would mean either duplicating the test file under a `dst_arrowstream` cfg, or — better — introducing a destination-parameterized test harness so both destinations run the same assertions. That is a separate refactor and out of scope here. (The `connectorx-python` `arrow_stream` tests are likewise currently limited to DECIMAL, so this is a pre-existing structural gap rather than a regression.)

The `arrowstream` mapping changes are identical in shape to the `arrow` ones and depend on the same type-system additions, so they are covered at the mapping level by the `arrow` tests; the only uncovered part is destination-specific wiring, which the harness above would address. Happy to follow up with that in a separate PR.

### DIFF
--- a/connectorx/src/destinations/arrow/arrow_assoc.rs
+++ b/connectorx/src/destinations/arrow/arrow_assoc.rs
@@ -8,9 +8,9 @@ use crate::{
 };
 use arrow::array::{
     ArrayBuilder, BooleanBuilder, Date32Builder, Decimal128Builder, Float32Builder, Float64Builder,
-    Int16Builder, Int32Builder, Int64Builder, LargeBinaryBuilder, LargeListBuilder, StringBuilder,
-    Time64MicrosecondBuilder, Time64NanosecondBuilder, TimestampMicrosecondBuilder,
-    TimestampNanosecondBuilder, UInt16Builder, UInt32Builder, UInt64Builder,
+    Int16Builder, Int32Builder, Int64Builder, Int8Builder, LargeBinaryBuilder, LargeListBuilder,
+    StringBuilder, Time64MicrosecondBuilder, Time64NanosecondBuilder, TimestampMicrosecondBuilder,
+    TimestampNanosecondBuilder, UInt16Builder, UInt32Builder, UInt64Builder, UInt8Builder,
 };
 use arrow::datatypes::Field;
 use arrow::datatypes::{DataType as ArrowDataType, TimeUnit};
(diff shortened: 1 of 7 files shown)


wangxiaoying:
Thanks @oystersuki !

Issue #677, opened 2024-08-30
Oracle -> PanicException: Type LongRaw not implemented for oracle!


GuillaumePressiat:
#### What language are you using?

Python

#### What version are you using?

connectorx-0.3.3-cp311-cp311-manylinux_2_28_x86_64.whl (60.5 MB)

#### What database are you using?

Oracle

#### What dataframe are you using?

polars

#### Can you describe your bug?

Panic exception when collecting an Oracle LONG RAW column datatype.

#### What are the steps to reproduce the behavior?

Table schema (data type LONG RAW)

```sql
CREATE TABLE SCHEM.TABL
(
  ID_FILE          NUMBER(15)   NOT NULL,
  FILE_CONTENT     LONG RAW
)
```

##### Example query / code

```python
import polars as pl
sql_query = """
select file_content from schem.tabl
"""
df = pl.read_database_uri(sql_query, db_uri)
```

#### What is the error?

``` 
thread '<unnamed>' panicked at /__w/connector-x/connector-x/connectorx/src/sources/oracle/typesystem.rs:53:18:
not implemented: Type LongRaw not implemented for oracle!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
---------------------------------------------------------------------------
PanicException                            Traceback (most recent call last)
File /opt/conda/lib/python3.11/site-packages/polars/io/database/_utils.py:54, in _read_sql_connectorx(query, connection_uri, partition_on, partition_range, partition_num, protocol, schema_overrides)
     53 try:
---> 54     tbl = cx.read_sql(
     55         conn=connection_uri,
     56         query=query,
     57         return_type="arrow2",
     58         partition_on=partition_on,
     59         partition_range=partition_range,
     60         partition_num=partition_num,
     61         protocol=protocol,
     62     )
     63 except BaseException as err:
     64     # basic sanitisation of /user:pass/ credentials exposed in connectorx errs

File /opt/conda/lib/python3.11/site-packages/connectorx/__init__.py:386, in read_sql(conn, query, return_type, protocol, partition_on, partition_range, partition_num, index_col)
    384 try_import_module("pyarrow")
--> 386 result = _read_sql(
    387     conn,
    388     "arrow2" if return_type in {"arrow2", "polars", "polars2"} else "arrow",
    389     queries=queries,
    390     protocol=protocol,
    391     partition_query=partition_query,
    392 )
    393 df = reconstruct_arrow(result)

[...]

Issue #547, opened 2023-09-29
Unable to read SQL Table with _date array: thread '<unnamed>' panicked at 'not implemented: _date',


Buedenbender:
#### What language are you using?

 **Python**

#### What version are you using?

[Replace this text with the version.](connectorx:          0.3.2)

#### What database are you using?

PostgresSQL

#### What dataframe are you using?

e.g. Pandas, Arrow

#### Can you describe your bug?

I would like to query an SQL (Postgres) table that contains a _date columns (array of dates).
I use polars for the query (https://github.com/pola-rs/polars/issues/11400)

#### What are the steps to reproduce the behavior?

Create exemplary data in SQL (postgres)
```sql
create table test_polars (
		arr_date _date
);

insert into test_polars 
values
		(array ['2023-01-01'::date,'2023-03-01'::date])
;

select * from test_polars;
```

Try to read data with [read_database](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.read_database.html)

```python
import polars as pl

pl.read_database("select * FROM test_polars",url)
```

#### What is the error?

```shell
thread '<unnamed>' panicked at 'not implemented: _date', /__w/connector-x/connector-x/connectorx/src/sources/postgres/typesystem.rs:108:22
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: <connectorx::sources::postgres::typesystem::PostgresTypeSystem as core::convert::From<&postgres_types::Type>>::from
   3: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   4: <connectorx::sources::postgres::PostgresSource<P,C> as connectorx::sources::Source>::fetch_metadata
   5: connectorx::dispatcher::Dispatcher<S,D,TP>::run
   6: connectorx::get_arrow2::get_arrow2
   7: connectorx::arrow2::write_arrow
   8: connectorx::read_sql::read_sql
   9: std::panicking::try
  10: connectorx::__pyo3_raw_read_sql
  11: <unknown>
  12: _PyEval_EvalFrameDefault
  13: _PyFunction_Vectorcall
  14: _PyEval_EvalFrameDefault
[...]

Issue #476, opened 2023-03-27
Panic when fetching 0000-00-00 date from MySQL Aurora


dafydds:
#### What language are you using?

Python 3

#### What version are you using?

connectorx==0.3.1
polars==0.16.14

#### What database are you using?

MySQL **AURORA**

#### What dataframe are you using?

Polars

#### Can you describe your bug?

Give a high level description of the bug.

#### What are the steps to reproduce the behavior?

If possible, please include a **minimal simple** example including:

##### Database setup if the error only happens on specific data or data type

MySQL AURORA

```
CREATE TABLE Person (
  id int(11) UNSIGNED NOT NULL AUTO_INCREMENT,
  birthDate date DEFAULT NULL
)

ENGINE = INNODB,
AUTO_INCREMENT = 1227146,
AVG_ROW_LENGTH = 180,
CHARACTER SET utf8,
COLLATE utf8_general_ci;
```

##### Example query / code

```
conn= "mysql://" + sql_username + ":" + sql_password + "@" + sql_host + ":" + sql_port + "/" + sql_main_database
sql = "select id, birthDate from Person where id = 1"  # this person has birthDate '0000-00-00'
df = pl.read_sql(sql, conn)  ## this player has a birth date of '0000-00-00'
```

#### What is the error?

    thread '<unnamed>' panicked at 'Could not retrieve core::option::Option<chrono::naive::date::NaiveDate> from Value', /github/home/.cargo/registry/src/github.com-1ecc6299db9ec823/mysql_common-0.27.5/src/value/convert/mod.rs:175:23
    Traceback (most recent call last):
    File "<stdin>", line 1, in <module>
    File "/home/ds/dev/work/prototypes/benh/run_sql/run_sql.py", line 180, in run_query
        df = pl.read_sql(query_str, conn_str)
    File "/home/ds/miniconda3/lib/python3.9/site-packages/polars/utils/decorators.py", line 35, in wrapper
        return function(*args, **kwargs)
    File "/home/ds/miniconda3/lib/python3.9/site-packages/polars/utils/decorators.py", line 126, in wrapper
        return function(*args, **kwargs)
    File "/home/ds/miniconda3/lib/python3.9/site-packages/polars/io/database.py", line 198, in read_sql
        return read_database(
    File "/home/ds/miniconda3/lib/python3.9/site-packages/polars/io/database.py", line 95, in read_database
        return _read_sql_connectorx(
[...]

Issue #268, opened 2022-04-12
' panicked at 'Could not retrieve core::option::Option<chrono::naive::datetime::NaiveDateTime> from Value'


ztsweet:
- the connectorx version is : 0.2.4

wangxiaoying:
Hi @ztsweet , we might need more information:

* What database are you using? (e.g. postgres, mysql)
* What `return_type` do you get? (e.g. pandas, arrow)
* Would be better if you could provide a simple example that we could reproduce your error

ztsweet:
- the code like:
`
df = pl.read_sql(sql, conn)
`
- the database is mysql
- I guess the most likely reason is that the datetime means like '000/00/00 00:00:00'

wangxiaoying:
Hi @ztsweet , sorry for the late reply. It seems like '0000/00/00 00:00:00' is not a valid datetime for mysql.

<img width="1102" alt="image" src="https://user-images.githubusercontent.com/5569610/166127887-e29d318b-1b5d-4a8b-ba3f-8566fd845cf1.png">

https://dev.mysql.com/doc/refman/8.0/en/datetime.html#:~:text=MySQL%20retrieves%20and%20displays%20DATETIME,both%20date%20and%20time%20parts.

kesmit13:
I get a similar error for date values that come back from the database as `0000-00-00`.

NamSahng:
I had a similar issue, when tried to select datetime columns on MySQL.
In my case, qeurying datetime column to string worked such as 
`SELECT DATE_FORMAT(date, '%Y-%m-%d') FROM TABLE`.

showkeyjar:
same error, and try catch not work.

sunshicheng:
I had a similar issue, when tried to select something on MySQL.
pyo3_runtime.PanicException: Could not retrieve chrono::naive::datetime::NaiveDateTime from Value
I use connectorx==0.3.1 ,first demo not work.

malachibazar:
I'm having the same issue on `connectorx==0.3.2` and a MySQL database.

Error: `panicked at 'Could not retrieve core::option::Option<chrono::naive::time::NaiveTime> from Value'`

Here's the scheme of the table:

![image](https://github.com/sfu-db/connector-x/assets/68713544/13afaed0-cb84-4dc1-ba9f-ac7653861f48)

malachibazar:
[...]

PR #920, opened 2026-06-24
fix: Rust panic in arrow_stream is not propagate to python

### LINKED ISSUE #913: Rust panic in arrow_stream is not propagated to Python

**Python**

0.4.5

MySQL

Arrow Stream

I encountered an issue when using `return_type="arrow_stream"` to read data from a MySQL table containing a DATETIME(6) column.

ConnectorX prints the following panic:

```
thread '<unnamed>' panicked at .../arrow_assoc.rs:290:28:
out of range DateTime
```

However, the panic is not propagated back to Python. Instead:

- No exception is raised.
- The process continues running normally.
- The panic message is only printed to stderr.

There are some odd DATETIME(6) values in the table (for example, 0001-04-17 21:12:06.000000). However, the same query succeeds when using `return_type="arrow"`. My main concern is that the error is not propagated as a Python exception when using `arrow_stream`; instead, the stream silently returns no batches while the Rust panic is only printed to stderr.

My expectation is that either:

`arrow_stream` should successfully read the column, as `arrow` does, or
the Rust panic should be converted into a Python exception instead of silently returning an empty stream.

| Field | Type |
|--------|--------|
| updated_at | datetime(6) |

```
0001-04-17 21:12:06.000000
2026-04-17 21:12:07.000000
```

##### Example query / code

``` python
import connectorx as cx
import polars as pl

print("Starting", flush=True)
stream = cx.read_sql(
    "mysql://user:password@localhost:3310/my_database",
    "SELECT updated_at FROM my_table LIMIT 10",
    return_type="arrow_stream",
    batch_size=1,
)

for batch in stream:
    df = pl.from_arrow(batch)
    print(df)

print(f"Finished", flush=True)
```

#### What is the error?

```
Starting
Finished

thread '<unnamed>' (66321) panicked at /Users/runner/work/connector-x/connector-x/connectorx/src/destinations/arrowstream/arrow_assoc.rs:290:28:
out of range DateTime
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Process finished with exit code 0
```

oystersuki:
## Summary

This fixes #913. When using return_type="arrow_stream", a Rust panic in the producer thread is not propagated to Python: no exception is raised, the process exits successfully, and the panic is only printed to stderr. The stream silently returns a truncated (often empty) result, which is particularly dangerous for data pipelines that then ingest incomplete data. The non-streaming arrow path does not have this problem because it already propagates the panic as a PanicException.

## Root cause

ArrowBatchIter::run() spawns the producer thread with std::thread::spawn(...) but does not keep the returned JoinHandle, so the thread is detached and its termination is never observed. When a partition panics, rayon re-raises the panic on that detached thread and the default hook prints "thread '<unnamed>' panicked at ...", but nothing ever reaches the consumer. In addition, next() used record_batch().ok().flatten(), which discards any error and maps it to None, making an early channel close indistinguishable from a normal end-of-stream.

## Fix

Store the JoinHandle on the iterator and join it when the stream reaches its end. If the producer panicked, re-raise the panic on the consumer thread (the thread calling into Python) with std::panic::resume_unwind, so PyO3 converts it into a PanicException at the __next__ boundary. If the producer returned a transport error, fail loudly as well. This makes arrow_stream consistent with the arrow path.

## Why the fix is value-agnostic

The silent path is fed by several unrelated panic sources: the nanosecond overflow on out-of-range datetimes reported in #913 (separately mitigated by the microsecond-precision change), zero dates such as 0000-00-00 from MySQL (#476, #268), and unsupported types (#547, #677). Because the fix surfaces the failure at the join boundary rather than handling specific values, every producer-side panic is propagated regardless of its cause.

## Scope and follow-up

This PR fixes propagation only. A transport error that is modeled as a Result is currently surfaced through panic! (and therefore as a PanicException) because Iterator::next can only return Option. Surfacing such errors as typed Python exceptions, matching the arrow path exactly, would require changing the RecordBatchIterator::next_batch signature to return a Result and deciding on a core error type. Because that changes the exception contract that downstream pipelines depend on, it is intentionally left out of this PR and proposed as a follow-up.

## Testing

Added a test using a dummy transport that panics in the middle of the stream on a specific value, and asserts via catch_unwind that the panic now propagates to the consumer instead of being silently swallowed.

### DIFF
--- a/connectorx/src/arrow_batch_iter.rs
+++ b/connectorx/src/arrow_batch_iter.rs
@@ -32,6 +32,7 @@ where
     dorder: DataOrder,
     src_schema: Vec<S::TypeSystem>,
     dst_schema: Vec<ArrowStreamTypeSystem>,
+    handle: Option<std::thread::JoinHandle<Result<(), TP::Error>>>,
     _phantom: PhantomData<TP>,
 }
 
@@ -61,6 +62,7 @@ where
             dorder,
             src_schema,
             dst_schema,
+            handle: None,
             _phantom: PhantomData,
         })
     }
@@ -72,7 +74,7 @@ where
         let dst_partitions = self.dst_parts.take().unwrap();
         let dorder = self.dorder;
 
-        std::thread::spawn(move || -> Result<(), TP::Error> {
+        self.handle = Some(std::thread::spawn(move || -> Result<(), TP::Error> {
             let schemas: Vec<_> = src_schema
                 .iter()
                 .zip_eq(&dst_schema)
@@ -132,7 +134,7 @@ where
             debug!("Writing finished");
 
             Ok(())
-        });
+        }));
     }
 }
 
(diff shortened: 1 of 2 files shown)


wangxiaoying:
Thank you @oystersuki , the code looks good to me.

PR #922, opened 2026-07-09
Add support for mysql SSL mode

### LINKED ISSUE #347: Add SSL certificate support for mysql
Add SSL certificate support for mysql

jessegrabowski:
Closes https://github.com/sfu-db/connector-x/issues/347

Currently connector-x does not support server-side SSL certification `?ssl-mode=REQUIRED`, and it is thus not possible to connect to a server that forces TLS. This functionality is available in the `native-tls` backend that connector-x's MySQL stack already uses, so this PR mostly just adds some parsing and argument forwarding.

I punted on full client-certificate mutual TLS, because that would be a deeper change and it wasn't requested in the linked issue. I also don't need it for my personal use-case, so there you go.

AI disclosure: This PR was written with the assistance of Claude Code, with me in the loop the whole time. I have reviewed the code manually and checked it against the contributing documentation I could find.

### DIFF
--- a/connectorx-python/connectorx/tests/test_mysql.py
+++ b/connectorx-python/connectorx/tests/test_mysql.py
@@ -1,10 +1,23 @@
+import os
+
 import pandas as pd
+import pytest
 from pandas.testing import assert_frame_equal
 
 from .. import read_sql, ConnectionUrl
 
 # mysql_url fixture is now defined in conftest.py
 # It uses testcontainers if available, otherwise the MYSQL_URL environment variable
+
+
+@pytest.fixture(scope="module")
+def mysql_url_tls() -> str:
+    return os.environ["MYSQL_URL_TLS"]
+
+
+@pytest.fixture(scope="module")
+def mysql_rootcert() -> str:
+    return os.environ["MYSQL_ROOTCERT"]
 
 
 def test_mysql_without_partition(mysql_url: str) -> None:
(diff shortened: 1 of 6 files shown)


connectorx-python/connectorx/tests/test_mysql.py
```diff
@@ -653,3 +666,62 @@ def test_mysql_decimal_pandas_unchanged(mysql_url: str) -> None:
...
+    reason="Do not test MySQL TLS unless `MYSQL_URL_TLS` is set",
+)
+def test_mysql_tls_required(mysql_url_tls: str) -> None:
+    df = read_sql(f"{mysql_url_tls}?ssl-mode=REQUIRED", "SELECT * FROM test_table")
+    df.sort_values(by="test_int", inplace=True, ignore_index=True)
+    assert_frame_equal(df, _expected_test_table(), check_names=True)
+
+
+@pytest.mark.skipif(
+    not os.environ.get("MYSQL_URL_TLS"),
```
oystersuki: This test requires two fixtures, but the skipif condition only specifies MYSQL_URL_TLS. In this case, if only MYSQL_ROOTCERT is missing, the test will not be skipped, resulting in a Key Error. Therefore, it would be best to add skipif to MYSQL_ROOTCERT as well.

oystersuki:
@jessegrabowski 

Once the comment is addressed and the conflicts are resolved, this should be good to merge!
Note that the 4 new tests don't run in CI yet, but I verified they all pass locally without any issues.
Going forward, we'll need to add a TLS-enabled MySQL container so these tests run in CI. I'll follow up on that separately.

jessegrabowski:
Thanks for taking a look at this. I rebased the branch onto main and updated the skip condition. Let me know if there's anything else you'd like to see, or if you want help with any follow-ups needed for a release.

oystersuki:
@jessegrabowski 


Thanks for the quick turnaround! LGTM!
I do have merge permissions as an external collaborator, but I don't handle the release process itself. The maintainers decide when to cut new releases.

That said, once this is merged, it should ride along with the next stable release by default—assuming no breaking changes sneak in before then. I'll just ping the maintainers to cut 0.4.6 stable once the current alpha2 validation period wraps up, so fingers crossed the timing works out.

PR #928, opened 2026-08-18
fix(trino): propagate URL connection parameters and fix partition verify bug

### LINKED ISSUE #927: Trino: partition queries ignore verify=false and connection URL parameters are not forwarded to client

Rust and Python (via `polars.read_database_uri`)

ConnectorX 0.4.6 (current main branch)

Trino

Arrow / Polars

The Trino source does not propagate URL query parameters to the `prusto` ClientBuilder. Specifically:

1. **`verify=false` is ignored in partition queries** — `trino_get_partition_range()` in `partition.rs` creates its own client without passing `no_verify`, causing TLS certificate errors on self-signed clusters when using `partition_on`.

2. **Session properties, extra credentials, and other Trino protocol headers are silently dropped** — the `prusto` ClientBuilder supports `.properties()`, `.extra_credentials()`, `.source()`, `.schema()`, `.client_tags()`, `.trace_token()`, and `.client_info()`, but ConnectorX never reads these from the connection URL.

This makes it impossible to set query timeouts, pass connector credentials (e.g. S3 tokens via `extra_credential`), route queries to resource groups (via `client_tags`), or set a default schema without qualifying every table name.

##### Example query / code

```python
import polars as pl

# Bug 1: verify=false ignored on partition query (TLS error on self-signed cluster)
df = pl.read_database_uri(
    "SELECT * FROM schema.table",
    "trino+https://user:pass@trino-selfsigned:443/hive?verify=false",
    engine="connectorx",
    partition_on="id",
    partition_num=4,
)
# The main query respects verify=false, but the SELECT MIN/MAX partition
# range query fails with a TLS certificate error.

# Bug 2: no way to pass session properties or extra credentials
df = pl.read_database_uri(
    "SELECT * FROM schema.table",
    "trino+https://user:pass@host:443/hive?session.query_max_execution_time=10m",
    engine="connectorx",
)
# The session property is silently ignored — Trino never receives it.
```

#### What is the error?

Bug 1 (partition + self-signed): TLS certificate verification failure on the `SELECT MIN(col), MAX(col)` pre-query, even though `verify=false` is in the URL.

Bug 2 (session properties): No error — parameters are silently ignored. The Trino server never receives `X-Trino-Session`, `X-Trino-Extra-Credential`, or other headers, which can cause queries to timeout with default settings or fail due to missing credentials.

---

I have a fix ready and would be happy to submit a PR if this is something you'd accept.

GiacomoSaccaggi:
## Summary

Fixes #927.

The Trino source drops all URL query parameters beyond basic auth, host, port, catalog, and `verify`. The `prusto` ClientBuilder already has methods for session properties, extra credentials, client tags, source, schema, etc. — they just were never called from ConnectorX.

On top of that, `trino_get_partition_range()` constructs its own client without passing `no_verify`, so `verify=false` gets lost on the `SELECT MIN/MAX` pre-query path. This causes TLS failures on self-signed clusters when using `partition_on`.

## Fix

Extracted a shared `build_client_from_url()` helper that wires up all supported `prusto::ClientBuilder` options from the URL. Both `TrinoSource::new` and `trino_get_partition_range` now call it — the partition bug is fixed by construction since both paths get identical configuration.

### Newly supported URL parameters

| Parameter | Header | Use case |
|---|---|---|
| `source=X` | `X-Trino-Source` | App identification in query logs |
| `schema=X` | `X-Trino-Schema` | Default schema |
| `client_tags=a,b` | `X-Trino-Client-Tags` | Resource group routing |
| `client_info=X` | `X-Trino-Client-Info` | Query metadata |
| `trace_token=X` | `X-Trino-Trace-Token` | Distributed tracing |
| `session.K=V` | `X-Trino-Session: K=V` | Session properties (timeouts, join type) |
| `extra_credential.K=V` | `X-Trino-Extra-Credential: K=V` | Connector credentials (S3, Hive) |

## Example

```python
import polars as pl

df = pl.read_database_uri(
    "SELECT * FROM events",
    "trino+https://user:pass@host:443/catalog?"
    "schema=analytics&source=my-pipeline&"
    "session.query_max_execution_time=10m&"
    "extra_credential.hive.s3.aws-access-key=AKIA...",
    engine="connectorx",
    partition_on="event_id",
    partition_num=8,
)
```

## Compatibility

Fully backward compatible — URLs without the new parameters behave exactly as before. The only observable change is that `partition_on` on self-signed clusters (with `verify=false`) no longer fails.

## Testing

- `cargo build --features src_trino` ✓
- `cargo fmt --all -- --check` ✓
- `cargo clippy --features src_trino` ✓ (no new warnings)
- 3 unit tests added (`test_new_with_all_params`, `test_new_minimal_url`, `test_new_ignores_empty_keys`)
- Verified with Polars 1.43.2 via `maturin develop --release`

## Write-up

https://giacomosaccaggi.github.io/deep-dives/connectorx_trino_params.html

### DIFF
--- a/connectorx/src/partition.rs
+++ b/connectorx/src/partition.rs
@@ -515,31 +515,12 @@ fn bigquery_get_partition_range(conn: &Url, query: &str, col: &str) -> (i64, i64
 #[cfg(feature = "src_trino")]
 #[throws(ConnectorXOutError)]
 fn trino_get_partition_range(conn: &Url, query: &str, col: &str) -> (i64, i64) {
-    use prusto::{auth::Auth, ClientBuilder};
-
-    use crate::sources::trino::{TrinoDialect, TrinoPartitionQueryResult};
+    use crate::sources::trino::{build_client_from_url, TrinoDialect, TrinoPartitionQueryResult};
 
     let rt = Runtime::new().expect("Failed to create runtime");
 
-    let username = match conn.username() {
-        "" => "connectorx",
-        username => username,
-    };
-
-    let builder = ClientBuilder::new(username, conn.host().unwrap().to_owned())
-        .port(conn.port().unwrap_or(8080))
-        .ssl(prusto::ssl::Ssl { root_cert: None })
-        .secure(conn.scheme() == "trino+https")
-        .catalog(conn.path_segments().unwrap().last().unwrap_or("hive"));
-
-    let builder = match conn.password() {
-        None => builder,
-        Some(password) => builder.auth(Auth::Basic(username.to_owned(), Some(password.to_owned()))),
-    };
-
-    let client = builder
-        .build()
-        .map_err(|e| anyhow!("Failed to build client: {}", e))?;
+    let client =
+        build_client_from_url(conn).map_err(|e| anyhow!("Failed to build Trino client: {}", e))?;
 
     let range_query = get_partition_range_query(query, col, &TrinoDialect {})?;
     let query_result = rt.block_on(client.get_all::<TrinoPartitionQueryResult>(range_query));
--- a/connectorx/src/sources/trino/mod.rs
+++ b/connectorx/src/sources/trino/mod.rs
@@ -1,4 +1,4 @@
-use std::{marker::PhantomData, sync::Arc};
+use std::{collections::HashMap, marker::PhantomData, sync::Arc};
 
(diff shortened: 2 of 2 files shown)


wangxiaoying:
Thanks @GiacomoSaccaggi , the code looks good to me!

PR #929, opened 2026-08-18
fix(sql): avoid query execution for schema discovery (LIMIT 0 + SQLite stmt.columns)

### LINKED ISSUE #768: Use `LIMIT 0` instead of `LIMIT 1` to get result schema.
This would probably save a lot of query execution time for queries which take a very long time to return limit 1 (e.g. with a subquery involving a SORT).

GiacomoSaccaggi:
## Summary

Fixes #768 — Use `LIMIT 0` instead of `LIMIT 1` to get result schema, avoiding expensive query execution.

### All sources except SQLite and Oracle
- Renamed `limit1_query` → `limit0_query` — returns only column metadata without reading any data rows.

### Oracle
- Uses `WHERE 1=0` (Oracle does not support `LIMIT`).

### SQLite
- Uses `conn.prepare()` + `stmt.columns()` to read `decl_type()` directly from the DDL — **no query execution at all**.
- Falls back to `LIMIT 1` only when `decl_type` is unavailable (computed expressions, aggregates like `COUNT(*)`, `AVG()`), where at least one row is needed to infer the type.
- This is strictly better than both `LIMIT 0` (which broke SQLite type inference) and `LIMIT 1` (which executed the full query plan).

## Improved behavior: empty result sets

Previously, queries returning zero rows would degrade all column types to `Text/object`. Now, schema discovery via `stmt.columns()` preserves the correct types even on empty results:

| Column | Before | After |
|--------|--------|-------|
| INTEGER | object | Int64 |
| REAL | object | float64 |
| BOOLEAN | object | boolean |
| DATE/DATETIME | object | datetime64[us] |
| TEXT | object | object |

Tests `test_empty_result` and `test_empty_result_on_partition` updated to reflect this improvement.

## Why not LIMIT 0 for SQLite?

SQLite's `query_row` with `LIMIT 0` returns `QueryReturnedNoRows`, providing no type information. The `stmt.columns()` approach gives schema metadata from the prepared statement without executing.

## Testing

- All existing Python SQLite tests pass (basic queries, partitions, LIMIT, aggregates, CTE, empty results, GROUP BY)
- All other sources compile and work unchanged with `limit0_query`
- `cargo check/clippy/fmt --features all` clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ This allows it to make full use of the CPU by becoming cache and branch predicto
 
 ## How does ConnectorX download the data?
 
-Upon receiving the query, e.g. `SELECT * FROM lineitem`, ConnectorX will first get the schema of the result set. Depending on the data source, this process may envolve issuing a `LIMIT 1` query `SELECT * FROM lineitem LIMIT 1`.
+Upon receiving the query, e.g. `SELECT * FROM lineitem`, ConnectorX will first get the schema of the result set. Depending on the data source, this process may involve issuing a `LIMIT 0` query `SELECT * FROM lineitem LIMIT 0`.
 
 Then, if `partition_on` is specified, ConnectorX will issue `SELECT MIN($partition_on), MAX($partition_on) FROM (SELECT * FROM lineitem)` to know the range of the partition column.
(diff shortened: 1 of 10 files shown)


wangxiaoying:
Thanks @GiacomoSaccaggi for the PR!

PR #933, opened 2026-08-31
fix(url): stop form-encoding connection URL parameters, and report unknown Postgres options as errors

GiacomoSaccaggi:
## Problem

Connection URL parameters whose value contains a space are corrupted before they reach the driver. With connectorx 0.4.5:

```python
import connectorx as cx

# ask the server for the application_name it sees
cx.read_sql(
    "postgresql://postgres:postgres@localhost:5432/testdb?application_name=my%20app",
    "SELECT application_name FROM pg_stat_activity WHERE pid = pg_backend_pid()",
)
# -> 'my+app'
```

The same substitution makes session settings unusable, which is the most common reason to put parameters in the URL at all:

```python
cx.read_sql(
    "postgresql://...?options=-c%20statement_timeout%3D1234ms",
    "SELECT current_setting('statement_timeout')",
)
# RuntimeError: db error: FATAL: unrecognized configuration parameter "+statement_timeout"
```

`%20`, `+` and a literal space all end up as `+`, so there is no way to express a space today. This is what pola-rs/polars#24488 runs into: the reporter has to strip `?options=-c%20default_transaction_read_only%3DTrue` from the URL when switching from `adbc` to `connectorx`.

Separately, a parameter that upstream rust-postgres does not recognize aborts the process:

```python
cx.read_sql("postgresql://...?totally_unknown_param=x", "SELECT 1")
# pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value:
#   Error { kind: ConfigParse, cause: Some(UnknownOption("totally_unknown_param")) }
```

## Cause

Two places rewrite the URL by rebuilding its query with `Url::query_pairs_mut().append_pair()`. That method serializes as `application/x-www-form-urlencoded`, where a space is written as `+`. Drivers that percent-decode the query — rust-postgres among them — then read that `+` literally.

- `SourceConn::try_from` (`source_router.rs`), when removing the `cxprotocol` parameter. This is the entry point for **every** source, so the bug is not Postgres-specific.
- `strip_bad_opts` (`sources/postgres/connection.rs`), when removing `sslkey` / `sslcert` / `sslrootcert`.

Both matter: fixing only one leaves the other to re-introduce the substitution. I verified that by reverting each site in turn and watching the new tests fail.

## Fix

A single helper, `remove_query_params`, filters the raw query segments and leaves the surviving parameters byte-for-byte as the caller wrote them, instead of decoding and re-encoding them. Both call sites use it. No new dependency: `form_urlencoded` is already re-exported by `url`, and is used only to compare parameter keys in decoded form, exactly as `query_pairs()` did.

`rewrite_tls_args` now propagates the `ConfigParse` error instead of calling `unwrap()`. `PostgresSourceError` already has `#[from] postgres::Error`, so this is a one-character change plus a test.

### This is not a behaviour change for callers who write `+`

| consumer | caller writes | before | after |
| --- | --- | --- | --- |
| percent-decoding (rust-postgres) | `%20` or a literal space | `+` | space |
| percent-decoding | `+` | `+` | `+` |
| form-decoding (`query_pairs`) | `%20` | space | space |
| form-decoding | `+` | space | space |

A caller who wrote `+` already got a literal `+`, because the old round-trip decoded `+` to a space and then re-encoded it back to `+`. Only values that were meant to contain a space change, and only from broken to correct. Parameters are now also passed through with the caller's own percent-encoding rather than a normalized one.

## Tests

Four new unit tests that need no database, so they run everywhere:

- `source_router::tests::keeps_remaining_query_params_verbatim` — `?options=-c%20statement_timeout%3D1s&cxprotocol=cursor` keeps `%20` and still picks up the protocol. Fails on `main` with `options=-c+statement_timeout%3D1s`.
- `source_router::tests::leaves_the_query_alone_when_there_is_nothing_to_remove` — also fails on `main`.
- Seven `utils::tests::*` covering the helper: spaces preserved, no re-encoding of `+`/`%2F`, multiple keys removed, query dropped when empty, missing query, percent-encoded keys, duplicate keys.

Three new integration tests in `tests/test_postgres.rs`, asserting the effect **on the server** rather than the shape of the string:

- `url_param_with_space_reaches_the_server` — `?application_name=cx%20e2e` and `pg_stat_activity` reports `cx e2e`. On `main`: `cx+e2e`.
- `url_options_param_is_applied_by_the_server` — `?options=-c%20statement_timeout%3D1234ms` and `current_setting('statement_timeout')` returns `1234ms`. On `main` the connection fails outright.
- `unknown_url_param_is_reported_as_an_error` — `rewrite_tls_args` returns `Err` instead of panicking.

Existing suites, run locally with the repository's testcontainers harness:

| suite | result |
| --- | --- |
| `test_postgres` | 21 passed, 0 failed |
| `test_mysql` | 28 passed, 0 failed |
| `test_trino` (including the `#[ignore]` ones) | 2 passed, 0 failed |
| `--lib` unit tests | 11 passed, 0 failed |

`cargo fmt --all -- --check` is clean, and `cargo clippy` reports nothing new on the touched files.

## Checked through the Python bindings too

I built the wheel from this branch and compared it against 0.4.5 on the same PostgreSQL 17 instance:

| case | 0.4.5 | this branch |
| --- | --- | --- |
| value containing a space | `'my+app'` | `'my app'` |
| `options=-c statement_timeout=1234ms` | `FATAL: unrecognized configuration parameter "+statement_timeout"` | `'1234ms'` |
| unrecognized parameter | `PanicException` | `RuntimeError: invalid connection string: unknown option ...` |

## What this unlocks for Polars

Polars is adding a `connection_options` parameter to `read_database_uri` (pola-rs/polars#29054), which merges a dict into the connection URI as query parameters for the connectorx engine. Values that contain a space — including the whole `options=-c name=value` family, which is how session settings are set — cannot work until this fix ships. With a wheel from this branch:

```python
pl.read_database_uri(
    "SELECT current_setting('statement_timeout')",
    "postgresql://postgres:postgres@localhost:5432/testdb",
    engine="connectorx",
    connection_options={"options": "-c statement_timeout=1234ms"},
)  # -> '1234ms'; on 0.4.5 the connection fails
```

That is also the scenario in pola-rs/polars#24488: one base URI reused across engines, with the extra parameters supplied separately, rather than stripped by hand for connectorx.

Related to pola-rs/polars#24488 and pola-rs/polars#29054.

### DIFF
--- a/connectorx/src/source_router.rs
+++ b/connectorx/src/source_router.rs
@@ -1,5 +1,6 @@
 use crate::constants::CONNECTORX_PROTOCOL;
 use crate::errors::{ConnectorXError, Result};
+use crate::utils::remove_query_params;
 use anyhow::anyhow;
 use fehler::throws;
 #[cfg(feature = "src_postgres")]
@@ -41,16 +42,7 @@ impl TryFrom<&str> for SourceConn {
         };
 
         // create url by removing connectorx protocol
-        let stripped_query: Vec<(_, _)> = old_url
-            .query_pairs()
-            .filter(|p| &*p.0 != CONNECTORX_PROTOCOL)
-            .collect();
-        let mut url = old_url.clone();
-        url.set_query(None);
-        for pair in stripped_query {
-            url.query_pairs_mut()
-                .append_pair(&pair.0.to_string()[..], &pair.1.to_string()[..]);
-        }
+        let url = remove_query_params(&old_url, &[CONNECTORX_PROTOCOL]);
 
         // users from sqlalchemy may set engine in connection url (e.g. mssql+pymssql://...)
         // only for compatablility, we don't use the same engine
@@ -92,4 +84,36 @@ pub fn parse_source(conn: &str, protocol: Option<&str>) -> SourceConn {
         None => {}
     }
     source_conn
+}
+
+#[cfg(test)]
+mod tests {
+    use super::SourceConn;
+    use std::convert::TryFrom;
+
+    /// Removing the connectorx protocol must not re-encode the remaining parameters:
+    /// sources that percent-decode the query would otherwise receive a literal `+`
+    /// wherever the caller wrote a space.
+    #[test]
+    fn keeps_remaining_query_params_verbatim() {
+        let source_conn = SourceConn::try_from(
+            "postgresql://u:p@host:5432/db?options=-c%20statement_timeout%3D1s&cxprotocol=cursor",
+        )
+        .unwrap();
+
+        assert_eq!(
+            source_conn.conn.query(),
+            Some("options=-c%20statement_timeout%3D1s")
+        );
+        assert_eq!(source_conn.proto, "cursor");
+    }
+
+    /// The query is left alone even when there is no protocol parameter to remove.
+    #[test]
+    fn leaves_the_query_alone_when_there_is_nothing_to_remove() {
+        let source_conn = SourceConn::try_from("mysql://host:3306/db?a=x%20y&b=%2Fz").unwrap();
+
+        assert_eq!(source_conn.conn.query(), Some("a=x%20y&b=%2Fz"));
+        assert_eq!(source_conn.proto, "binary");
+    }
 }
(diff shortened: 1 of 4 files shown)


wangxiaoying:
Thank you @GiacomoSaccaggi !

PR #934, opened 2026-09-04
fix(oracle): map unconstrained NUMBER(0,0) to Decimal instead of float64

dbascoules:
An Oracle `NUMBER` with no declared precision/scale (e.g. the result of `SUM()` or an arithmetic expression) was mapped to `NumFloat`, silently losing precision beyond f64's ~15-17 significant digits.

This routes it through the existing `NumDecimal`/`rust_decimal::Decimal` path instead, consistent with the `Number(*,>0)` case just below it.

Validated against a live Oracle instance: `SUM()` and arithmetic expressions over NUMBER columns now come back as exact decimal128 instead of float64.

### DIFF
--- a/connectorx/src/sources/oracle/typesystem.rs
+++ b/connectorx/src/sources/oracle/typesystem.rs
@@ -40,7 +40,7 @@ impl<'a> From<&'a OracleType> for OracleTypeSystem {
     fn from(ty: &'a OracleType) -> OracleTypeSystem {
         use OracleTypeSystem::*;
         match ty {
-            OracleType::Number(0, 0) => NumFloat(true),
+            OracleType::Number(0, 0) => NumDecimal(true),
             OracleType::Number(_, 0) => NumInt(true),
             OracleType::Number(_, _) => NumDecimal(true),
             OracleType::Float(_) => Float(true),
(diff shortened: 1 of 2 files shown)


dbascoules:
For context: this brings Oracle's unconstrained `NUMBER` in line with how other connectors handle the equivalent case.

- **PostgreSQL** ([postgres/typesystem.rs](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/postgres/typesystem.rs#L100)): `numeric` always maps to `Decimal`, regardless of declared precision/scale — Postgres doesn't even distinguish on that at the type-dispatch level.
- **MySQL** ([mysql/typesystem.rs](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/mysql/typesystem.rs#L115-L116)): `DECIMAL`/`NEWDECIMAL` → `Decimal`.
- **MSSQL** ([mssql/typesystem.rs](https://github.com/sfu-db/connector-x/blob/main/connectorx/src/sources/mssql/typesystem.rs#L88-L89)): `Decimaln`/`Numericn` → `Decimal`.
- **ClickHouse**: `Decimal(precision, scale)` → `Decimal`.

Trino and BigQuery are the exceptions (decimal/numeric → `f64`), but they apply that uniformly to *all* decimal columns, constrained or not — they don't have an "unconstrained" special case like Oracle's `NUMBER(0,0)`, so it's not really comparable to the bug fixed here.

wangxiaoying:
Thanks @dbascoules for the PR, it looks good to me!

PR #935, opened 2026-09-05
fix(python): regenerate poetry.lock and pin Poetry version

oystersuki:
The release workflow on `prerelease` fails on `apple-arm (3.14t)` during `just bootstrap-python-release`.

`poetry install --only main --no-root` pulls in polars, dask, pandas and pyarrow, even though all four are `optional = true` in `pyproject.toml`. On 3.14t there's no matching polars wheel (polars ships abi3 only, and the stable ABI isn't available on free-threaded builds), so it falls back to building polars from source and dies there. Other Python versions are fine because the abi3 wheel installs directly, which is why this only showed up once 3.14t was added.

### Cause

`poetry.lock` was generated by Poetry 2.1.4, which doesn't write the per-extra `markers` entries for optional deps. Poetry 2.3.0 flipped the default of `installer.re-resolve` from `true` to `false` (python-poetry/poetry#10622), so newer Poetry no longer re-resolves against `pyproject.toml` at install time and trusts the lock alone. Without the markers, everything in the main group looks required.

`Gr1N/setup-poetry@v9` installs the latest stable release when `poetry-version` isn't set, so CI drifted past 2.3 while `.devcontainer` still pins 2.1.3. That's why nobody could reproduce it locally.

### Verified

macOS arm64, CPython 3.14.7+freethreaded, same lock file, clean venv each time — only the Poetry version differs:

| Poetry | `poetry install --only main --no-root` |
| --- | --- |
| 2.1.3 | 1 install (numpy) |
| 2.2.1 | 1 install (numpy) |
| 2.3.0 | 17 installs |
| 2.4.1 | 17 installs |
| 2.4.1 + `installer.re-resolve=true` | 1 install (numpy) |

### Changes

- Regenerated `poetry.lock` with 2.4.1 so the markers are written
- Bumped `POETRY_VERSION` in `.devcontainer/Dockerfile` to 2.4.1
- Pinned `poetry-version: "2.4.1"` on all four `Gr1N/setup-poetry` steps

The regenerated lock installs fine under both 2.1.3 and 2.4.1, so it isn't tied to a specific version. The devcontainer and workflow pins are there because running `poetry lock` with 2.1.3 strips the markers straight back out.

### Not in this PR

- The published wheels have no dependency metadata at all — `0.4.5` and `0.4.6a1` have zero `Requires-Dist`/`Provides-Extra`, so `pip install connectorx` doesn't pull numpy and `connectorx[polars]` is a silent no-op. maturin reads `[project]`, which has no `dependencies` or `optional-dependencies`.
- `maturin-version` is pinned to `v0.14.15` on `win-and-mac` and `apple-arm`. Building cp314t wheels needs a current maturin — I built `connectorx-0.4.6a1-cp314-cp314t-macosx_11_0_arm64.whl` locally with 1.14+ in ~3 min, so bumping it should be enough to get apple-arm (3.14t) green.
- `verify` has `needs: [win-and-mac, linux]` but its matrix includes `macos-latest`, whose artifact comes from `apple-arm`.
- 3.14t wheels alone aren't free-threading support — the module needs `gil_used = false`, and the partition-thread GIL handling should be looked at first.


oystersuki:
@wangxiaoying 

This should be why the release run failed the other day — `bootstrap-python-release` was building polars from source on 3.14t and fail on the maturin build.

Targeting `prerelease` rather than `main` since that's where the workflow actually runs and where the stale lock lives. Happy to redo it against `main` if you'd rather it go the other way.

wangxiaoying (approved):
Thanks @oystersuki !

PR #940, opened 2026-09-11
feat: change mysql test image

oystersuki:
The current tests use ghcr.io/wangxiaoying/mysql, a fork of official mysql:8.0.25. Checking it with docker history shows its CMD only adds defaults (utf8mb4) on top of `mysqld:
CMD ["mysqld", "--init-connect='SET NAMES utf8mb4'", "--skip-character-set-client-handshake", "--collation-server=utf8mb4_unicode_ci", "--character-set-server=utf8mb4"]`

This PR swaps to the official mysql:8.0.25 and passes those same settings as command arguments, so behavior is equivalent. TLS cannot be enabled on the wangxiaoying image (SSL_CTX_set_default_verify_paths failed), which blocks the TLS tests added in #922 from running in CI. The official image auto-provisions self-signed certificates on startup, so the #922 tests can be executed.

Verified with the full MySQL test suite (Python and Rust) producing the same results before and after. A follow-up PR will wire the TLS test fixtures into conftest.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -26,7 +26,12 @@ services:
       - connectorx-network
 
   mysql:
-    image: ghcr.io/wangxiaoying/mysql:latest
+    image: mysql:8.0.25
+    command:
+      - --character-set-server=utf8mb4
+      - --collation-server=utf8mb4_unicode_ci
+      - --skip-character-set-client-handshake
+      - --init-connect=SET NAMES utf8mb4
     environment:
       MYSQL_DATABASE: mysql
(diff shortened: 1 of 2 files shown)


oystersuki:
Sorry to bother you when you are busy, but could @wangxiaoying please take a look at this since it's a change to the test image? Thank you in advance.

wangxiaoying:
Thanks @oystersuki , the mysql image indeed was a workaround at that time and should be updated with the official one!

oystersuki:
Thank you! I was thinking that at the time, building a custom image was probably necessary because GitHub Actions didn't support custom commands back then. But since we've now switched to Testcontainers, I figure it's no longer needed!

PR #943, opened 2026-09-13
Fix Databricks import abort from bundled Kerberos/OpenSSL dependencies

### LINKED ISSUE #831: cant import library due to FIPS issue in databricks container

**Python**.

3.12.11, connectorx 0.4.3

PostgreSQL, but irrelevant

Polars, but irrelevant

Importing connectorx in a docker image built from databricksruntime/minimal:16.4-LTS (dockerfile here: https://github.com/databricks/containers/blob/master/ubuntu/minimal/Dockerfile) gives a FIPS SSL error:

```
import connectorx
crypto/fips/fips.c:154: OpenSSL internal error: FATAL FIPS SELFTEST FAILURE
```

This causes the program to abort.

Just trying to build this image shows the error for me:
```
# minimal.Dockerfile
FROM databricksruntime/minimal:16.4-LTS

RUN apt-get update && apt-get install -y git wget curl
ADD https://astral.sh/uv/install.sh /uv-installer.sh
RUN sh /uv-installer.sh

# Ensure the installed binary is on the `PATH`
ENV PATH="/root/.local/bin/:$PATH"

RUN uv venv
RUN uv pip install connectorx
RUN uv run python -c "import connectorx"
```

```
docker build -t myimage:latest -f minimal.Dockerfile .
[+] Building 13.5s (11/11) FINISHED                                                                                                                                                     docker:default
 => [internal] load build definition from minimal.Dockerfile                                                                                                                                      0.0s
 => => transferring dockerfile: 386B                                                                                                                                                              0.0s
 => [internal] load metadata for docker.io/databricksruntime/minimal:16.4-LTS                                                                                                                     0.5s
 => [internal] load .dockerignore                                                                                                                                                                 0.0s
 => => transferring context: 2B                                                                                                                                                                   0.0s
 => [1/7] FROM docker.io/databricksruntime/minimal:16.4-LTS@sha256:d1092be719617a6a89cc343fc75ef69936944737f3315315a56ad09e908cfe22                                                               0.0s
 => => resolve docker.io/databricksruntime/minimal:16.4-LTS@sha256:d1092be719617a6a89cc343fc75ef69936944737f3315315a56ad09e908cfe22                                                               0.0s
 => [3/7] ADD https://astral.sh/uv/install.sh /uv-installer.sh                                                                                                                                    0.1s
 => CACHED [2/7] RUN apt-get update && apt-get install -y git wget curl                                                                                                                           0.0s
 => CACHED [3/7] ADD https://astral.sh/uv/install.sh /uv-installer.sh                                                                                                                             0.0s
 => CACHED [4/7] RUN sh /uv-installer.sh                                                                                                                                                          0.0s
 => [5/7] RUN uv venv                                                                                                                                                                             6.0s
 => [6/7] RUN uv pip install connectorx                                                                                                                                                           6.3s
 => ERROR [7/7] RUN uv run python -c "import connectorx"                                                                                                                                          0.5s 
------
 > [7/7] RUN uv run python -c "import connectorx":
0.482 crypto/fips/fips.c:154: OpenSSL internal error: FATAL FIPS SELFTEST FAILURE
------

 1 warning found (use docker --debug to expand):
 - InvalidBaseImagePlatform: Base image databricksruntime/minimal:16.4-LTS was pulled with platform "linux/amd64", expected "linux/arm64" for current build (line 1)
minimal.Dockerfile:12
--------------------
  10 |     RUN uv venv
  11 |     RUN uv pip install connectorx
  12 | >>> RUN uv run python -c "import connectorx"
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c uv run python -c \"import connectorx\"" did not complete successfully: exit code: 134
```

N/A

##### Example query / code

N/A

#### What is the error?

Shown above

Let me know if I can help with anything else, thank you for the awesome work of this repo

saurabh500:
## Summary

Fixes #831.

Linux wheels currently bundle a distribution-patched OpenSSL 1.1 library through Kerberos/GSSAPI. Loading that library alone aborts in the reported Databricks image, even with its inherited `OPENSSL_FORCE_FIPS_MODE=0`. This is separate from Rust's vendored OpenSSL dependency and occurs before any database connection.

- Build checksum-pinned MIT Kerberos 1.22.2 against OpenSSL 3 in the existing manylinux 2.28 build, retaining GSSAPI and including redistribution notices.
- Preserve the manylinux runtime plugin directories (`/usr/lib64/krb5/plugins`, including legacy loaders, and `/usr/lib64/gss`). Stage installation away from system files and keep linker search paths pointed at the rebuilt libraries, so wheel repair does not pick up old system Kerberos.
- Run upstream crypto/GSSAPI checks and check the linked native dependencies.
- Check relative plugin registration, explicit `plugin_base_dir` overrides, missing-plugin behavior, repeated use, and legacy loader paths against the installed wheel in every Linux build before artifact upload.
- Add fresh-process wheel checks for repeated imports, both import orders with Python `ssl`, unchanged crypto environment, and optional repeated SQLite-to-Polars queries.
- Require verification in the pinned Databricks 16.4-LTS image before publication. Add a manual Linux x86_64/Python 3.12 build-and-check path that does not publish, and a published-wheel Databricks check.
- Document the failure and compatibility scope. Plugins are not bundled; applications must supply compatible plugins and their dependencies.

No Python API, Rust runtime code, or Cargo dependency changes. Host FIPS settings and TLS verification are not altered. This is an import-compatibility fix, not a claim of FIPS-compliant or certified operation.

## Validation

- Reproduced the native abort with ConnectorX 0.4.3, Python 3.12.11, and the exact Linux amd64 image digest from the issue. Published 0.4.6a1 also fails.
- Isolated the failure to the bundled `libcrypto.so.1.1`; removing the environment trigger only in a diagnostic control makes the old import succeed.
- Confirmed the plugin regression: published 0.4.5 finds a controlled relative hostrealm plugin; the original PR wheel fails; explicitly restoring the old base directory makes it succeed. Confirmed the published 0.4.5 x86_64 and ARM64 wheels use the same plugin directory defaults.
- Built a fresh local CPython 3.12/manylinux 2.28 x86_64 wheel from source with the final native build recipe and GSSAPI enabled. SHA-256: `f657a87758facc195ae8a6f6ff806d16b993c61a94cc851b2506add3ffe1b682`. This locally built 0.4.6a2 candidate is not a downloaded or published release artifact.
- Final wheel contains rebuilt Kerberos, shared OpenSSL 3, and both redistribution notices, with no bundled OpenSSL 1.1. Native dependency resolution also selects the rebuilt libraries with build environment overrides removed.
- Plugin checks pass on the final wheel: default lookup and a distinct explicit override, each queried twice on each of two contexts. Removing each fixture makes lookup fail; a missing override does not fall back to the default directory. Legacy authdata, service-locator, and GSS mechanism paths are checked in the packaged binaries.
- The regression rejects both the original PR wheel and an intermediate build that preserved modern lookup but missed legacy directories. Full-wheel verification also rejected an intermediate build that picked up old system libraries during wheel repair.
- Ran the entire plugin regression twice in clean manylinux 2.28 with neither system OpenSSL 3 nor the build prefix installed. Fresh-process imports, both `ssl` orders, and two SQLite-to-Polars queries also pass there; process mappings confirm the wheel's bundled OpenSSL 3 is loaded.
- The final wheel passes the committed Databricks Docker wrapper with Python 3.12.11 and inherited `OPENSSL_FORCE_FIPS_MODE=0`. The regression explicitly rejects a missing/changed trigger, wrong Python version, or wrong architecture.
- Upstream MIT Kerberos crypto and GSSAPI checks pass with the final recipe. Checksums confirm staged installation leaves existing system plugin files unchanged. Earlier native validation recorded three hostname-canonicalization skips because the container hostname had no alias.
- ShellCheck for the build script, Python formatting, and whitespace checks pass. Workflow structure passes actionlint with existing obsolete-action/runner diagnostics excluded; existing workflow ShellCheck findings are outside this change, and none were reported on the added step.

## Remaining coverage / review focus

- ARM, Windows, and macOS builds were not run locally.
- Live SQL Server/Active Directory authentication and database TLS integration were not exercised. Runtime plugin loading was exercised with the controlled hostrealm module; legacy loader and GSS mechanism paths received binary checks, not third-party plugin integration tests.
- GitHub-hosted workflow execution remains to be observed; validation above was local.
- No package release has been published.


wangxiaoying:
Hi @saurabh500 , thank you for the contribution and it could enable an important use case for us!

The change looks good to me. Only one question: is there a specific reason that the python version is 3.12? If the import error also happens in later python versions, will it be better to have the verification workflow running on it (e.g., 3.14)?

saurabh500:
@wangxiaoying The fix was limited to Python 3.12 because that is the floor of the Py version in the issue. The same fix can be applied to all downlevel python versions too. Want me to make the change?

wangxiaoying:
> @wangxiaoying The fix was limited to Python 3.12 because that is the floor of the Py version in the issue. The same fix can be applied to all downlevel python versions too. Want me to make the change?

Hi @saurabh500 , thanks for the quick reply. I think we can keep it like this for now.

Issue #460, opened 2023-03-06
Can not support to connect apache doris or starrocks through mysql protocol


ulnit:
#### What language are you using?

ROAPI

#### What version are you using?

Replace this text with the version.

#### What database are you using?

Apache Doris

#### What dataframe are you using?

  Arrow

#### What is the error?

WARN connectorx::sources::mysql] mysql text prepared statement error: MySqlError { ERROR 1064 (HY000): Unsupported command(COM_STMT_PREPARE) }, switch to limit1 method
Error: Database error: MySqlError { ERROR 1064 (HY000): Unsupported command(COM_STMT_PREPARE) }

l1xnan:
I had the same problem, it should occur during the prepared phase:
https://github.com/sfu-db/connector-x/blob/7b3147436b7e20b96691348143d605e2249d6119/connectorx/src/sources/mysql/mod.rs#L101

It might have something to do with it: https://github.com/StarRocks/starrocks/pull/41080

PR #947, opened 2026-09-17
fix(mysql): read fallback schema from LIMIT 0 result set metadata

oystersuki:
## Summary

Fixes a regression from #929 (in 0.4.6a2) in the MySQL source's schema-discovery fallback.

When `COM_STMT_PREPARE` fails, `fetch_metadata` runs the query with a `LIMIT` and derives the column types from the result. #929 changed that `LIMIT 1` to `LIMIT 0`, but the MySQL caller still expected a row (`query_first` → `Ok(Some(row))`). `LIMIT 0` never returns one, so the fallback always ended in the last-resort branch that marks every column `VarChar(false)`:

- numeric / decimal / datetime columns come back as strings
- `NULL` or binary values panic (`Could not retrieve alloc::string::String`)

## Fix

Read the column definitions from the `LIMIT 0` result set (`QueryResult::columns()`) instead of from a row. MySQL sends them even when the result set is empty, so no data row is read, and the schema is now also correct for queries that return zero rows — something the old `LIMIT 1` code could not do. The unreachable "all columns VarChar" branch is removed. `limit0_query` and the other sources are untouched.

## Affected setups

Servers that speak the MySQL protocol but reject server-side prepared statements, used with `protocol="text"`: SingleStore with audit logging, Apache Doris / StarRocks (#460), proxies that don't forward `COM_STMT_PREPARE`. Regular MySQL / MariaDB never take the fallback and are unaffected.

## Reproduction

```sql
SET GLOBAL max_prepared_stmt_count = 0;  -- makes COM_STMT_PREPARE fail
```

```python
cx.read_sql(url, "SELECT id, amt, ts FROM t", protocol="text", return_type="arrow").schema
# before: id: string, amt: string, ts: string  (panics if the table has NULLs or VARBINARY)
# after:  id: int32,  amt: decimal128(38, 10), ts: timestamp[us]
```

## Tests

- New `test_mysql_schema_fallback_without_server_side_prepare`: disables preparing via `max_prepared_stmt_count = 0` (restored afterwards), then checks that the fallback schema and data match the prepared-statement path for both a non-empty and an empty result. Fails before the fix, passes after. If the restore itself fails (only possible if the fallback is broken), the test says so and prints the manual restore command instead of letting later tests fail confusingly.
- Existing Rust and Python MySQL tests pass; `cargo fmt` / `cargo clippy` clean.


oystersuki:
Hi @wangxiaoying

I said 0.4.6a2 felt ready for a full release — I need to walk that back a bit. While testing further I found a regression in 0.4.6a2 that my earlier testing didn't cover: since #929 (LIMIT 1 → LIMIT 0), the MySQL source's schema-discovery fallback always reports every column as string. It only affects servers where COM_STMT_PREPARE fails (e.g. SingleStore with audit logging, Apache Doris/StarRocks as in #460, or proxies), used with protocol="text", which is why
my MySQL 5–8 runs never hit it. On those setups 0.4.5 works and 0.4.6a2 degrades
numeric/date columns to strings or panics on NULL / binary values.

The fix is small (MySQL source only, other sources unaffected) and comes with a regression test: #<PR>. Could this go into 0.4.6 before the full release? I'd rather not have a release that breaks these users when 0.4.5 works for them.

Sorry for the noise, and thanks.

wangxiaoying:
@oystersuki no problem and thanks for pointing this out! This is now merged and will be included into the stable release.

PR #950, opened 2026-09-18
Fix Clippy warnings across the workspace

saurabh500:
## Summary

Clippy cleanup for the workspace, in two commits. No behavior change.

## 1. Unused imports and lifetimes

- **Unused imports** — removed 14 redundant `use crate::impl_transport;` / `use crate::impl_typesystem;` lines across `destinations/{arrow,arrowstream}/typesystem.rs` and all 12 `transports/*_arrow{,stream}.rs` files. Both macros are `#[macro_export]`'d, which makes them textually available crate-wide without an explicit `use`.
- **Unused lifetimes** — dropped the unused `'a` lifetime parameter from 14 `impl` blocks in `arrow_batch_iter.rs`, `sources/bigquery/mod.rs`, and `sources/bigquery/typesystem.rs`. In each case `'a` only appeared in a trivially-satisfiable `S: Source + 'a` where-bound (or not at all) and was never tied to the impl's `Self` type, so it was removed along with the now-unneeded `+ 'a` bound.

## 2. Remaining lint categories

`connectorx`:

| Lint | Where |
|---|---|
| `manual_map` | `destinations/{arrow,arrowstream}/arrow_assoc.rs` |
| `manual_ok_err` | `destinations/arrowstream/mod.rs` |
| `single_match` | `source_router.rs`, `sources/mssql/mod.rs` |
| `collapsible_match` | `sources/csv/mod.rs` |
| `needless_late_init` | `sources/csv/mod.rs`, `sql.rs` |
| `unnecessary_literal_unwrap` | `fed_rewriter.rs` — `InvocationArg::try_from(x).unwrap()` → `InvocationArg::from(x)` |
| `get_first` | `sources/clickhouse/typesystem.rs` |
| `double_ended_iterator_last` | `sources/trino/mod.rs` — `.last()` → `.next_back()` |
| `suspicious_to_owned` | `sources/mssql/mod.rs` |
| `dead_code` | `transports/dummy_arrowstream.rs` |

Two of these are worth a closer look:

**`suspicious_to_owned`** — `.to_owned()` on a `Cow<str>` clones the `Cow`, it does not produce a `String`. These are now `.into_owned()`. Note that `decode(url.username())?.to_owned()` appears in *two* `cfg` branches — one under `#[cfg(any(windows, feature = "integrated-auth-gssapi"))]` and one under the negation. Windows Clippy only flags the first; Linux CI would flag the second. Both are fixed.

**`dead_code`** — `dummy_arrowstream` was the only `*_arrowstream` transport never re-exported from `transports/mod.rs`, which is why its types looked dead. Fixed by adding the missing re-export with the same `...ArrowStream...` alias convention every other stream transport uses, rather than by adding an `allow`.

`connectorx-cpp`:

- **`missing_safety_doc` (14)** — added a real `# Safety` section to each `unsafe` fn covering pointer validity, NUL-termination, and single-free semantics, plus which function releases each returned allocation.
- **`arc_with_non_send_sync` (3)** — suppressed with `#[allow]` and a comment. These `Arc`s are FFI ownership handles: created, immediately `Arc::into_raw`'d for the C caller, and reclaimed by a matching `Arc::from_raw` in the free function. They are never cloned or moved across threads, so `FFI_ArrowArray`/`FFI_ArrowSchema` lacking `Send + Sync` does not matter here. `Box` would model the single ownership better, but that means changing six `from_raw` sites and the FFI ownership contract, which cannot be exercised from the Rust test suite. Flagging it as a deliberate choice rather than an oversight.
- `manual_ok_err` in `connectorx_rewrite`.

## Out of scope

- **`result_large_err` (37)** — caused by the size of the shared `ConnectorXOutError` enum in `errors.rs`, not by any individual function. Boxing it is a public API change and belongs in its own PR.
- `partition.rs`'s `get_first` and the `while_let_loop` in `arrowstream/mod.rs` are handled in #952.
- A handful of one-off lints remain (`unreachable_patterns`, `type_complexity`, `from_over_into`, `let_and_return`, and similar).

## Verification

Using the exact CI commands from `.github/workflows/ci.yml`:

- `cargo clippy --features all` — exit 0, `connectorx` lib warnings **430 → 371**
- `cargo clippy --features all --all` — exit 0, `connectorx-cpp` warnings **20 → 2**
- `cargo fmt --all -- --check` — exit 0

Every remaining `connectorx-cpp` warning and every remaining grouped `connectorx` warning was confirmed to be either `result_large_err`, an out-of-scope one-off, or owned by #952. No new warnings were introduced.


wangxiaoying:
Thank you, @saurabh500! I’ve invited you as a contributor. Hopefully this will make it easier for you to contribute to the project!

PR #953, opened 2026-09-18
fix(trino): decode URL credentials before authentication

npennequin:
## Summary

- Parse Trino connection URLs without decoding the entire URL first.
- Percent-decode the username and password before using them for Basic
  authentication.
- Reuse `build_client_from_url` in `TrinoSource::new`.
- Add regression tests for reserved characters, encoded usernames, and
  URLs without userinfo.

## Problem

`TrinoSource::new` percent-decoded the complete connection string before
parsing it as a URL.

For credentials containing reserved characters such as `@`, `/`, `:`,
`#`, `?`, or spaces, this either changed the URL structure or caused the
URL parser to encode the character again. The encoded value was then sent
literally in the Basic Authorization header, and Trino rejected otherwise
valid credentials with `401 Unauthorized`.

Encoding the password at the call site did not work around the issue,
because ConnectorX decoded the entire URL before parsing it.

## Solution

Parse the connection URL as-is, preserving its structure, and decode only
the username and password returned by `url::Url`.

`TrinoSource::new` now delegates client construction to the existing
`build_client_from_url` function. This also ensures that the regular
`read_sql` path honors the same Trino connection parameters as the
partitioned path, including `source`, `schema`, `client_tags`,
`trace_token`, session properties, and extra credentials.

## Testing

- `cargo test -p connectorx --no-default-features \
  --features src_trino,dst_arrow,fptr --lib sources::trino`
  - 6 tests passed.
- `cargo fmt --all -- --check`
- Full optimized Python extension build with `maturin develop --release`.
- End-to-end validation against a Trino cluster using a password
  containing `@`:
  - direct HTTP Basic authentication succeeded;
  - the Python `trino` client returned one row;
  - the patched ConnectorX build returned one row.


oystersuki:
Thanks @npennequin! 

the fix itself looks right. Two small requests before merging: the branch needs a rebase (main changed .last() to .next_back() in the block you removed, so take your side of the conflict), and I'd like build_client_from_url to match what the other hand-parsed sources do. MsSQL and Oracle decode() the path segment and use host_str().unwrap_or("localhost"); here the catalog is left percent-encoded (/my%5Fcatalog used to work, since the old code decoded the whole string) and url.host().unwrap() panics on a host-less URL.

↓like this
```
    let catalog = match url.path_segments().and_then(|mut s| s.next_back()) {
        Some(segment) => decode(segment)?.into_owned(),
        None => "hive".to_owned(),
    };

    let mut builder =
        ClientBuilder::new(&username, url.host_str().unwrap_or("localhost").to_owned())
            .port(url.port().unwrap_or(8080))
            .ssl(prusto::ssl::Ssl { root_cert: None })
            .no_verify(no_verify)
            .secure(url.scheme() == "trino+https")
            .catalog(&catalog);
```

npennequin:
Thank you for the review @oystersuki

I've rebased the branch and implemented your suggested changes.

Would you like me to squash the two commits?

oystersuki:
LGTM! Thanks for addressing the Codecov suggestions as well! No need to squash those two commits.

PR #954, opened 2026-09-18
fix(trino): decode BIGINT values as i64

npennequin:
## Summary

Fix Trino `BIGINT` values being incorrectly decoded through `i32` in the Arrow and Arrow-stream transports.

## Problem

The Trino type system correctly associates `BIGINT` with `i64`, and `serde_json` preserves signed 64-bit JSON integers exactly. However, both Arrow transport mappings requested `BIGINT` values from the source parser as `i32`.

Values outside the `i32` range therefore failed during narrowing, including `2518422941645303032`. This produced errors such as `Trino cannot parse i64 at position: (0, 9)`.

## Changes

- Map Trino `BIGINT` from `i64` to Arrow `Int64`.
- Apply the fix to both Arrow and Arrow-stream transports.
- Make the existing `INTEGER` widening from `i32` to `i64` explicit.
- Add regression tests using a mocked Trino JSON response.
- Cover the observed value, values beyond 2^53, and the signed `i64` boundaries.

## Validation

- Trino-focused Rust tests pass on Linux: 10 passed.
- Both `fptr` and `branch` transport paths pass locally.
- `cargo fmt` and diff checks pass.
- A release Python wheel was tested against a populated Trino/Hive partition using a full-schema query:
  - 1 row returned
  - All 62 columns returned
  - Both warm-up and timed reads succeeded


oystersuki:
Thanks @npennequin!

Just had two quick thoughts on tests and Pandas support.

Python regression tests:
connectorx/tests/test_trino.rs is currently #[ignore]d, so CI skips it. The Python suite actually runs against real Trino via Testcontainers, so it'd be safer to put the test there.
If we update test_types.test_bigint in scripts/trino.sql to full i64 bounds (-9223372036854775808 / 9223372036854775807) like our other DB seeds, and add an Arrow round-trip check in test_trino.py (similar to test_mysql.py), we'll catch this properly. Once that's added, we can clean up the two unit tests in sources/trino/mod.rs since they won't be needed anymore.

Pandas transport:
connectorx-python/src/pandas/transports/trino.rs is still set to Bigint[i32]. Since read_sql uses pandas by default, updating the SQL seed will break test_trino_types_binary. It should just be the same 2-line fix as the Arrow transport, so mind including that here as well?

npennequin:
Thank you @oystersuki for the review!

I've implemented both points.

oystersuki:
LGTM! 
Nit: to_pylist() is preferable over to_pandas().to_list(), but since the type is ensured by asserting == pa.int64(), it works fine as a boundary value test.  That said, to_pandas().to_list() is also used in Python tests for other data sources, so this isn't an issue.

PR #615, opened 2024-04-21
[WIP] support python 3.12

wangxiaoying:
#527

PR #956, opened 2026-09-19
test(partition): add unit tests for partition() and wire cargo test --lib into CI

saurabh500:
## Why
PR #899 changes `connectorx::partition::partition()` to guard against division-by-zero and integer overflow. This PR adds unit tests against `main` (pre-fix) so the current, well-defined behavior can be validated as unchanged once #899 lands, and so any intentional edge-case behavior changes are explicit/reviewed rather than silent regressions.

## What
1. Adds `#[cfg(test)] mod tests` in `connectorx/src/partition.rs` covering `partition()`: even/uneven ranges, single-partition, single-row, negative ranges, partially-specified range (error), large-but-safe range, many small partitions, and two baseline (pre-fix) behaviors — `num == 0` panics with "attempt to divide by zero", and `max < min` produces nonsensical but non-panicking bounds.

   Tests use a `SourceConn` backed by SQLite (`src_sqlite` feature), which lets `get_part_query` generate real SQL without a live DB connection, so bounds can be asserted directly on the generated SQL text.

2. **Adds `cargo test --lib --features all` to the `test-ci` recipe in `Justfile`.**

   `test-ci` only invokes specific integration test binaries via `--test <name>` (`test_postgres`, `test_polars`), which does not run `cargo test --lib` — so unit tests embedded in `src/*.rs` (these new `partition::tests`, plus pre-existing `source_router::tests` and `utils::tests`) haven't been executing in CI, only locally.

   `git blame` shows this was introduced deliberately in d06bd523 ("due to space limit on ci vm, only do selected tests"), landed via #615 by **@wangxiaoying**, to cut disk usage on the CI VM (integration tests pull Docker images via `testcontainers`). That correctly targeted the expensive container-backed tests, but also dropped unit test execution since `--lib` was never in the reduced target list. `cargo test --lib --features all` is container-free and fast, so it shouldn't reintroduce the disk pressure — tagging @wangxiaoying in case there's more context on the VM constraints.

## Validation
```
cargo test --lib --features src_sqlite partition::
cargo test --lib --features all partition::
cargo fmt -- --check
cargo clippy --lib --features src_sqlite --tests
```
All pass; no new clippy warnings introduced.


saurabh500:
Merged since the run didn't cause any failures in CI due to low diskspace.

PR #961, opened 2026-09-19
Phase 3: runtime MSSQL driver switch, flip default to mssql-tds

saurabh500:
Stacked on Phase 2.

## What this does

- Compiles Tiberius and `mssql-tds` into the same binary through an
  enum-dispatch source wrapper.
- Adds a process-wide runtime selector, defaulting new MSSQL sources to
  `mssql-tds`.
- Exposes `connectorx.mssql_driver` so Python wheel users can opt back into
  Tiberius without rebuilding:

  ```python
  import connectorx as cx
  cx.mssql_driver = "tiberius"
  ```

- Adds an end-to-end test that runs the same full type-matrix query through
  both drivers in one process and compares Arrow output.
- Documents the default change and compatibility switch.

This is the first customer-visible phase: new builds default to
`mssql-tds`; Tiberius remains available as a temporary fallback.

## Validation

- Rust live SQL Server tests pass under Tiberius-only, `mssql-tds`-only, and
  dual-driver configurations.
- The Python extension builds with `maturin develop --release`.
- The Python MSSQL suite passes under both runtime choices: 27 passed and
  one expected failure each time.


connectorx-python/src/lib.rs
```diff
@@ -35,12 +35,44 @@ fn connectorx(_: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
...
     m.add_class::<pandas::PandasBlockInfo>()?;
     m.add_class::<arrow::PyRecordBatch>()?;
     m.add_class::<arrow::PyRecordBatchIterator>()?;
     Ok(())
 }
 
+/// Returns the MSSQL driver ConnectorX currently uses: `"tiberius"` or
+/// `"mssql-tds"`. Backs the `connectorx.mssql_driver` module property; see
+/// sfu-db/connector-x#942 Phase 3.
+#[cfg(feature = "srcs")]
```
saurabh500: Why the name srcs ? This is not very readable

saurabh500: `srcs` means “sources”: it is the Python crate's existing Cargo feature that bundles database sources, paired with `dsts` for destinations. This PR reuses it because it enables both MSSQL drivers, which these getter/setter functions require. Added a clarifying comment at its definition in aeb28ca0f. I kept the existing name to avoid breaking builds using `--features srcs`; renaming the bundle would be a separate compatibility change.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

PR #957, opened 2026-09-19
ci: report Rust unit test coverage with Codecov

saurabh500:
## Why
#956 wired `cargo test --lib` into CI, but the workflow does not report how much code those tests exercise.

## What
1. Install `cargo-llvm-cov` in the existing Rust CI job and run the container-free `--lib` unit tests with `--features all`.
2. Generate `lcov.info`, write the overall coverage summary to the GitHub Actions job summary, and upload the report with `codecov/codecov-action@v5`.
3. Add `codecov.yml` with a 60% patch-coverage target, informational project coverage, and diff/file details in the PR comment.

This intentionally uses Codecov instead of maintaining a second privileged `workflow_run`, artifact download, `diff-cover`, and custom sticky-comment action. Codecov owns the fork-safe upload/reporting boundary and keeps this repository's CI configuration small.

## One-time onboarding
A `sfu-db` organization admin must install/authorize the Codecov GitHub App for `sfu-db/connector-x`, enable the repository in Codecov, and add its upload token as the `CODECOV_TOKEN` GitHub Actions repository secret. Until then, uploads are non-blocking (`fail_ci_if_error: false`) and no coverage comment/status will appear.

## Validation
- `.github/workflows/ci.yml` and `codecov.yml` parse successfully.
- `git diff --check` passes.

@wangxiaoying tagging you since you're familiar with the CI disk/VM constraints from #615; coverage remains scoped to the container-free unit tests.

PR #965, opened 2026-09-19
ci: stop running --lib tests twice for coverage

saurabh500:
## Mechanism
`just test-ci`'s `cargo test --lib --features all` and `cargo llvm-cov --lib --features all` compile and run the exact same `--lib` test suite, back to back. `cargo-llvm-cov` sets coverage-instrumentation `RUSTFLAGS`, which differ from a plain `cargo test` build and invalidate its build cache — so this was a full compile+run of the same tests twice per CI job, not a one-time coverage overhead.

## Fix
1. Make `cargo-llvm-cov` the only runner of the `--lib` suite.
2. Split unit and integration coverage into two reports instead of one combined `lcov.info`: `cargo llvm-cov --no-report` accumulates coverage counters without generating a report, and `cargo llvm-cov clean --workspace --profraw-only` resets those counters between the unit and integration runs *without* touching compiled artifacts (only `--lib`/library artifacts under the same instrumented build are removed and relinked, no plain/instrumented flag flip). So `lcov-unit.info` and `lcov-integration.info` each reflect only their own suite, uploaded to Codecov under separate `unit-tests`/`integration-tests` flags, while the whole run still shares a single instrumented compile.

New `Justfile` recipes: `coverage-unit` and `coverage-integration`.

## Validation
- `.github/workflows/ci.yml` and `Justfile` parse/format successfully.
- `git diff --check` passes.
- Confirmed via #957's just-merged CI run that "Run tests" (13 min, includes `--lib`) and "Generate unit test coverage" ran sequentially and would each rebuild/rerun `--lib`.


.github/workflows/ci.yml
```diff
@@ -74,7 +76,11 @@ jobs:
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
 
+      - name: Run integration tests
+        run: just test-ci-integration
+
       - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v5
```
saurabh500, line 88: This was added twice. Why ?

.github/workflows/ci.yml
```diff
@@ -74,7 +76,11 @@ jobs:
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
 
+      - name: Run integration tests
```
saurabh500: Is this new now ?

.github/workflows/ci.yml
saurabh500: Good catch, that was a stray duplicate line from an earlier edit, not intentional — removed in 2e8c5a91a.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

.github/workflows/ci.yml
saurabh500: Yes — this PR splits the old single `test-ci` step (which ran `--lib` unit tests + both integration test binaries together) into two steps: `cargo-llvm-cov` now runs `--lib` once for coverage, and this new step runs only the `test_postgres`/`test_polars` integration tests via the new `test-ci-integration` Justfile recipe, so `--lib` isn't built and run twice.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

Justfile
```diff
@@ -20,6 +20,26 @@ test-ci:
     cargo test --features src_postgres --features dst_arrow --test test_postgres
     cargo test --features src_postgres --features src_dummy --features dst_polars --test test_polars
 
+# integration tests only, for CI jobs that get --lib coverage from cargo-llvm-cov instead
+test-ci-integration:
```
saurabh500: Is this unused?

saurabh500: Good catch — yes, it was superseded by `coverage-integration` in the same PR but I forgot to remove it. Deleted in 5b808ec61.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

